PR #37, opened 2019-03-10
Fix NRE when custom controller implementation does not have a visualizer

wassx:
**Overview**

When implementing a custom controller without a visualizer (Keyboard controller) a NRE occurred in check when accessing GameObjectProxy

**Breaking Changes:**

- nothing

### DIFF
--- a/XRTK-Core/Assets/XRTK/Utilities/Solvers/ControllerFinder.cs
+++ b/XRTK-Core/Assets/XRTK/Utilities/Solvers/ControllerFinder.cs
@@ -100,7 +100,7 @@ namespace XRTK.SDK.Utilities.Solvers
         /// <param name="newController">The new controller to be tracked.</param>
         protected virtual void AddControllerTransform(IMixedRealityController newController)
         {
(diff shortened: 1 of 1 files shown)


XRTK-Core/Assets/XRTK/Utilities/Solvers/ControllerFinder.cs
```diff
@@ -100,7 +100,7 @@ protected virtual void TryAndAddControllerTransform()
         /// <param name="newController">The new controller to be tracked.</param>
         protected virtual void AddControllerTransform(IMixedRealityController newController)
         {
-            if (newController.ControllerHandedness == handedness && newController.Visualizer.GameObjectProxy.transform != null && !newController.Visualizer.GameObjectProxy.transform.Equals(ControllerTransform))
+            if (newController.ControllerHandedness == handedness && newController.Visualizer != null && newController.Visualizer.GameObjectProxy.transform != null && !newController.Visualizer.GameObjectProxy.transform.Equals(ControllerTransform))
```
StephenHodgson: I wonder if we also need to check if the GameObjectProxy is null as well 🤔

PR #44, opened 2019-03-26
Component extensions use == null instead unity's Object.bool

wassx:
**Overview**

Same in: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3691

**Changes:**

- Fixes: #...

**Breaking Changes:**

- Breaks...


wassx:
PR has wrong assumptions. Closing.

PR #47, opened 2019-03-26
Services that utilize uwp capabilities, now properly set them

StephenHodgson:
Ensures that services that use a uwp component are setting them when they're constructor is called in the editor, which then updates the player settings.

When the Mixed Reality Toolkit instance is created all services get their constructors called, even at edit time.

### DIFF
--- a/XRTK-Core/Assets/XRTK.WindowsMixedReality/Controllers/WindowsSpeechDataProvider.cs
+++ b/XRTK-Core/Assets/XRTK.WindowsMixedReality/Controllers/WindowsSpeechDataProvider.cs
@@ -37,6 +37,13 @@ namespace XRTK.WindowsMixedReality.Controllers
                 throw new Exception("Null speech commands in the speech commands profile!");
             }
 
+#if UNITY_WSA && UNITY_EDITOR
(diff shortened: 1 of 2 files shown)


StephenHodgson:
@Alexees I wouldn't mind your review on this, if you have time.

Alexees:
This seem to work, two thoughts though:
- Neither disabling the Spatial Awareness System in the Toolkit configuration profile nor removing it from there removes the capability again.
- This would lead to another fancyness consideration. If there was another service that needs this capability we would need to count them to toggle them off if not one of the services using the capability is active anymore.

StephenHodgson:
> Neither disabling the Spatial Awareness System in the Toolkit configuration profile nor removing the spatial mesh observer from there removes its capability again.

I'm not sure it matters if the capability needs to be removed if it's removed. I don't mind this being a manual task. No harm comes from keeping it unless it's a restricted capability.

> This would lead to another fancyness consideration

The current implementation or the idea you have? I was just going to force them to true, no record keeping there.

Alexees:
just the idea i had. if keeping them true if fine don't bother :)

Jarodshow:
Thanks!

PR #51, opened 2019-03-29
Boundary System Cleanup

StephenHodgson:
**Overview**

There was a duplicate enum that wasn't really being used in the boundary system and could be safely removed.

The `ExperienceScale` settings wasn't actually doing anything and the seated vs standing experience should be setup via the WMR Portal.

For applications that want to enforce a specific experience scale such as sitting or standing, then they should implement their own scripts that satisfy their needs and constraints. For the most part, we'll let the editor and platform determine the default experience scale.

**Breaking Changes:**

- Removed redundant `ExperienceScale` enum. Use `UnityEngine.XR.TrackingSpaceType` instead
- Removed `Scale` from `IMixedRealityBoundarySystem`
- Removed `TargetExperienceScale` from `MixedRealityToolkitConfigurationProfile`


SimonDarksideJ:
Looks good, but for historical documentation around the change, can we have a screenshot in the PR for the new screen, and a short sentence on how we envision the ExperienceScale / TrackingSpaceType would be managed per project?

PR #54, opened 2019-03-29
Unity Package Manager Project Changes

StephenHodgson:
**Overview**

Updates the project to the upm format needed to facilitate the ease of use with the Unity Package Manager. Each of the subfolders have already been moved out of the repository and into their own so that their respective upm packages can be easily built/released.

**TODO**
- [ ] Add XRTK Package specific window for adding additional packages via GitHub
- [ ] Make it easy to switch from package to local files for easier editing in a single project.
- [ ] Figure out how docs will be affected. (might need to generate docs in their own repo with the docs websites setup as subdomains, and xref them together)
- [x] Update the readme to show how to manually add upm package to manifest.json
- [x] Update the readme page to point to the respective repositories for each platform.
- [x] Ensure the other repos are setup for upm
- [x] Fix project pathfinding for assets/resources. (might end up replacing them with scriptable objects that reference the textures)image


StephenHodgson:
Note to self: Need to fix the upm extension version to 0.9.1

PR #57, opened 2019-04-02
Unity Package Manager Project Changes

StephenHodgson:
**Overview**

Updates the project to the upm format needed to facilitate the ease of use with the Unity Package Manager. Each of the subfolders have already been moved out of the repository and into their own so that their respective upm packages can be easily built/released.

**TODO**
- [ ] Figure out how docs will be affected. (might need to generate docs in their own repo with the docs websites setup as subdomains, and xref them together)
- [ ] Make it easy to switch from package to local files for easier editing in a single project.
- [x] Add XRTK Package specific window for adding additional packages via GitHub
image
- [x] Update the readme to show how to manually add upm package to manifest.json
- [x] Update the readme page to point to the respective repositories for each platform.
- [x] Ensure the other repos are setup for upm
- [x] Fix project pathfinding for assets/resources. (might end up replacing them with scriptable objects that reference the textures)image


SimonDarksideJ:
Looking good but hard to review with so many files touched. Are all these needed? as even examples have been touched.
Can we split out the non-essential file changes or refresh with dev to clear these?

StephenHodgson:
Mainly just ripping out all the other packages and putting them in new repositories.

`MixedRealityEditorSettings` and `MixedRealityPackageUtilities` are really the only two files that changed

SimonDarksideJ:
Then those SHOULD have been a separate cleanup PR for clarity :D 
Will check those two files now.

StephenHodgson:
Unfortunately I can't split up this PR without breaking the build bc then there'd be two copies of the each class.

SimonDarksideJ:
I suspect it's more a note for the future (and learning from our past experiences)
1. Move FIRST
2. THEN update :D

SimonDarksideJ:
The actual changes look fine, but they are still lost in the mist.

StephenHodgson:
> 1: I can uninstall the UPM GitExtension, should this be allowed?

Yes, it can be uninstalled but and I did consider keeping it from being uninstalled.

> 2: Does the install / uninstall also include dependencies, if so we should have a test / example (E.g. have Examples depend on the SDK ?) I tested by uninstalling both XRTK.wmr and Unity.WMR. When installing XRTK.wmr, the Unity package was not installed (and vice versa)

Yes it does but I didn't setup the dependencies on the other projects yet. that's not a blocker for this PR as it's something we need to address in those other packages.

> 3: Does not warn of missing dependencies (although this might be because dependencies aren't tested yet, 2 above). installing XRTK.WMR should warn if Unity.WMR is missing. same with XRTK.Examples missing XRTK.SDK

See 2

4: Uninstalling WMR causes a ton of issues in console. This categorically should not happen.

See 2

5: UI should be named as "Mixed Reality Packages" in title and in the menu, so it's not confused with the Unity package manager of the same name.

It's a menu directly under the Mixed Reality Toolkit, I think it's okay. 

6: Can we cache the package list and refresh in the background once open. If there are potentially a lot of packages, this UI would become very slow.

That's something we can iterate on when we get there. For now we'll only have a handful.  The plan is to better support the upm workflow by adding [scoped registries](https://forum.unity.com/threads/setup-for-scoped-registries-private-registries.573934)

StephenHodgson:
I fixed the dependencies in the other repositories to address issues 1, 2, & 3

StephenHodgson:
Actually, it seems I've broken it with adding the dependencies?

Maybe there's a different format Unity is expecting. Let me look into that.

XRTK-Core/Assets/XRTK/Inspectors/MixedRealityPackageUtilities.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        /// <summary>
+        /// Is the package utility running a check?
+        /// </summary>
+        public static bool IsRunningCheck { get; private set; } = false;
+
+        /// <summary>
+        /// Debug the package utility.
+        /// </summary>
+        public static bool DebugEnabled { get; set; } = true;
```
StephenHodgson:
```suggestion
        public static bool DebugEnabled { get; set; } = false;
```

PR #120, opened 2019-04-26
More robust cross platform support

StephenHodgson:
- Removed the profile check for the spatial awareness data provider's profiles, that should be handled in the constructor of the respective data provider with a thrown exception if missing.

- Moved the type null checks until after the platform check to better support false positive errors thrown when that type/assembly is probably missing because we're targeting a different platform.
- updated `TypeRefPropertyDrawer` to not throw an error if the type is missing because we're targeting a different platform.

### DIFF
--- a/XRTK-Core/Assets/XRTK/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
+++ b/XRTK-Core/Assets/XRTK/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Configuration;
 using UnityEditor;
 using UnityEditor.Compilation;
 using UnityEngine;
@@ -248,7 +247,6 @@ namespace XRTK.Inspectors.PropertyDrawers
                         if (!reference.Contains(" {missing}"))
                         {
                             reference += " {missing}";
-                            Debug.LogError($"Failed to resolve class reference for property '{property.name}' on {property.serializedObject.targetObject.name}");
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


XRTK-Core/Assets/XRTK/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
```diff
@@ -330,11 +328,11 @@ private static Type[] FindTypesByName(string typeName, SystemTypeAttribute filte
             var types = new List<Type>();
             var filteredTypes = GetFilteredTypes(filter);
 
-            for (var i = 0; i < filteredTypes.Count; i++)
+            foreach (var type in filteredTypes)
```
SimonDarksideJ: Why change a performant FOR loop for a more expensive ForEach?

XRTK-Core/Assets/XRTK/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
```diff
@@ -347,22 +345,22 @@ private static void DisplayDropDown(Rect position, List<Type> types, Type select
             menu.AddItem(new GUIContent("(None)"), selectedType == null, OnSelectedTypeName, null);
             menu.AddSeparator(string.Empty);
 
-            for (int i = 0; i < types.Count; ++i)
+            foreach (var type in types)
             {
-                string menuLabel = FormatGroupedTypeName(types[i], grouping);
+                var menuLabel = FormatGroupedTypeName(type, grouping);
```
SimonDarksideJ: Why the sudden change of explicit types to using var, we already know the format and type for each of these.

XRTK-Core/Assets/XRTK/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
StephenHodgson: I'm not worried about the for each loops in editor code.

XRTK-Core/Assets/XRTK/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
StephenHodgson: I think it's pretty obvious that it's a string from the context.

PR #175, opened 2019-05-09
EditorBuildSettings.scenes fixes

SimonDarksideJ:

## Overview

If accessed when not initialized (I.E. in a brand new Unity project) Unity will crash.
To avoid this, the necessary checks have been added to ensure it's ready before use to avoid this Unity bug.

## Target of the change:

Is this enhancement for:

## Changes:

- Fixes: Adds null checking and Length checking of the EditorBuildSettings.scene array


StephenHodgson:
Try running the build locally.

SimonDarksideJ:
Can't replicate but I've added some extra checking around the lines it's reporting an issue.

StephenHodgson:
Hmm. I wonder why the build is still failing?

SimonDarksideJ:
OK, I've reverted the additional changes beyond the original bug (the extra changes I made for protection).  If this still fails I'll redo the PR from a fresh branch

SimonDarksideJ:
Ok, so that build worked.  Will just have to merge this one fix and then re-test.
No idea why the other build failed which is concerning.

PR #181, opened 2019-05-12
Updated Awaiter.WaitUntil extension method with indefinite option

StephenHodgson:

## Overview

Needed the ability to wait indefinitely.

## Target of the change:

Is this enhancement for:

- Extension


StephenHodgson:
So by default this awaiter will timeout after 10 seconds, but there may be instances when people want to wait indefinitely. I didn't want to make it indefinite by default, nor break the existing functionality.

As far as use cases? I'm not sure what you mean. Like adding example code showing it?

SimonDarksideJ:
More like a laymans terms example for why it might be used (guess you have one else you wouldn't have added it._

StephenHodgson:
I'm not sure how I could simplify it anymore. Could you help?

afaik it just waits forever. (although in the future I'd love to add a cancellation token of some sort as an additional argument in an overload.

PR #195, opened 2019-05-19
Ports and formatting

StephenHodgson:

## Overview

- ported deadzone fixes
- enabled warnings as errors
- misc formatting updates

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Definitions/Utilities/MixedRealityPose.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Definitions/Utilities/MixedRealityPose.cs
@@ -122,7 +122,7 @@ namespace XRTK.Definitions.Utilities
         public bool Equals(MixedRealityPose other)
         {
             return Position == other.Position &&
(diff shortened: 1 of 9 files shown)


XRTK-Core/Packages/com.xrtk.core/Definitions/Utilities/MixedRealityPose.cs
```diff
@@ -122,7 +122,7 @@ bool IEqualityComparer.Equals(object left, object right)
         public bool Equals(MixedRealityPose other)
         {
             return Position == other.Position &&
```
SimonDarksideJ: Why not also use Vector2.Equals?

StephenHodgson: Because the equality operator is enough here.

The equality operator on the quaternion is prone to float precision errors, where as calling `Equals` outright does not.

This change was ported from https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4309

Issue #200, opened 2019-05-23
Type or namespace name 'ExecuteEvents' could not be found


hridpath:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug
Reciving the following error in Unity Console:
```
Failed to register the MixedRealityInputSystem service: System.Exception - The Input system is missing the required GazeProviderType!
  at XRTK.Services.InputSystem.MixedRealityInputSystem..ctor (XRTK.Definitions.InputSystem.MixedRealityInputSystemProfile profile) [0x000da] in O:\Git\Source\Repos\XRTK\XRTK-Core\XRTK-Core\Assets\XRTK\Services\InputSystem\MixedRealityInputSystem.cs:44 

  at (wrapper managed-to-native) System.Reflection.MonoCMethod.InternalInvoke(System.Reflection.MonoCMethod,object,object[],System.Exception&)

  at System.Reflection.MonoCMethod.InternalInvoke (System.Object obj, System.Object[] parameters) [0x00002] in <1f0c1ef1ad524c38bbc5536809c46b48>:0 
UnityEngine.Debug:LogError(Object)
XRTK.Services.MixedRealityToolkit:CreateAndRegisterService(Type, SupportedPlatforms, Object[]) (at Assets/XRTK/Services/MixedRealityToolkit.cs:888)
XRTK.Services.MixedRealityToolkit:CreateAndRegisterService(Type, Object[]) (at Assets/XRTK/Services/MixedRealityToolkit.cs:833)
XRTK.Services.MixedRealityToolkit:InitializeServiceLocator() (at Assets/XRTK/Services/MixedRealityToolkit.cs:383)
XRTK.Services.MixedRealityToolkit:InitializeInstance() (at Assets/XRTK/Services/MixedRealityToolkit.cs:282)
XRTK.Services.MixedRealityToolkit:get_Instance() (at Assets/XRTK/Services/MixedRealityToolkit.cs:206)
XRTK.Services.MixedRealityToolkit:ConfirmInitialized() (at Assets/XRTK/Services/MixedRealityToolkit.cs:319)
XRTK.Inspectors.Utilities.MixedRealityInspectorUtility:CheckMixedRealityConfigured(Boolean) (at Assets/XRTK/Inspectors/Utilities/MixedRealityInspectorUtility.cs:45)
[...]

PR #201, opened 2019-05-24
Unity 2019.2 Pre-Fixes

SimonDarksideJ:

## Overview

Pre-resolves an issue in 2019.2 whereby the UI system has been moved out of the editor in to a package.

Confirmed adding this reference (even though earlier versions don't recognise it) does not impact operation.

## Target of the change:

## Changes:

Adds an asmdef reference to the Unity.ugui package in the Core / Inspectors definitions

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Inspectors/XRTK.Inspectors.asmdef
+++ b/XRTK-Core/Packages/com.xrtk.core/Inspectors/XRTK.Inspectors.asmdef
@@ -4,7 +4,8 @@
         "XRTK",
         "XRTK.Extensions.EditorClassExtensions",
         "XRTK.Utilities.Build",
-        "XRTK.Utilities.Editor"
+        "XRTK.Utilities.Editor",
+        "Unity.ugui"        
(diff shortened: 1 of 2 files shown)


SimonDarksideJ:
Fixes towards #200

hridpath:
@Stephen Hodgson ref: #201; Can we get this change into a new UPM package release? I would like to test it. I want to see if I can get back to my original issue. TIA

SimonDarksideJ:
These fixes should resolve most of the critical issues, but I feel more testing/fixes will be needed once these are in a package.
But you can test locally @hridpath by cloning the XRTK-Core project, switching to development and pulling in these changes from the fix branch.  Be sure to also pull in the changes in the UpmGitExtension submodule too

StephenHodgson:
Are these changes backward compatible?

hridpath:
I believe that @SimonDarksideJ stated that Unity will ignore these if they do not exist.

hridpath:
I am not sure how best to test using local repro

PR #226, opened 2019-06-25
Input Simulation Port: Base

FejZa:
# XRTK - Mixed Reality Toolkit Change Request

## Overview

These are the base changes required to add input simuation to the XRTK. The simulation is based on the MRTK simulation with adjustments to the XRTK, suggested by me. Please also consider the related PR in SDK (https://github.com/XRTK/SDK/pull/46)

## Target of the change:

Is this enhancement for:

- Core (core framework, interfaces and definitions)

## Changes:

- Add input simulation data providers profile
- Add input simulation data providers inspector
- Add simulation data providers to input system profile
- Add simulation data providers to input system profile inspector
- Update `MixedRealityToolkit` to register simulation data providers

PR #227, opened 2019-06-25
Input Simulation Port: Camera

FejZa:

## Overview

This PR builds on top of #226 and adds camera input simulation. The user can navigate the scene using the keyboard and simulate head motion. Please also consider the related PR in SDK https://github.com/XRTK/SDK/pull/47.

## Target of the change:

Is this enhancement for:

## Changes:

- Define inspector and profile for camera input simulation
- Define camera input simulation data provider


StephenHodgson:
I'm not sure simulation should be used in conjunction with keyboard.

It should be a default input method for platforms that support keyboard input.

FejZa:
I am not sure I understand. How are you going to move / simulate without a keyboard?

StephenHodgson:
There's going to be a need for a switch to disable typical input and switching to the simulated.

This probably just needs more discussion more than anything,

FejZa:
So you mean there needs to be a mechanism e.g. to detect if a headset is connected and providing input then disable the camera simulation?

FejZa:
Which makes me aware of an issue I need to fix. All the simulation must only run when in Editor. Forgot about that.

SimonDarksideJ:
For the editor, input should support Keyboard / Mouse / Gamepad.   (Possibly leap motion for hands, but that is another PR)
Other than that, it isn't relevant for the editor environment.

Headsets are definitely not included, as that would use it's own native input and simulation would not be required

PR #230, opened 2019-07-05
Remove unused profile from BaseControllerDataProvider

FejZa:

## Overview

Remove unused profile parameter required by `BaseControllerDataProvider` to increase flexibility.

## Target of the change:

Is this enhancement for:

- Platform Lumin / WMR

## Changes:

- Removed the parameters from the base constructor and updated data providers to omit it


FejZa:
The reason why I'd remove this unused parameter is that it limits the usage of the base to data providers that use `BaseMixedRealityControllerDataProviderProfile`, which leads to some difficulties when implementing e.g. a "`SimulatedHandTrackingControllerDataProvider`", which has a different profile base.

StephenHodgson:
Why doesn't `SimulatedHandTrackingControllerDataProvider` have the same base?

Issue #233, opened 2019-07-08
Controller Visualisation not behaving as expected


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

The original intention with the Controller Visualisation Configuration, was that it provided ways to OVERRIDE the default configuration of the Visualiser.

In such, with nothing configured, it would visualise the Gizmo models (the model defaults) for al platforms, regardless of whether any overrides were configured.  However, today it only renders a model if an override is applied.

## To Reproduce

1. Create new XRTK project and configure
2. View Visualizer configuration and remove any configured overrides
3. Run the XRTK project with a headset / controllers connected

## Expected behavior

With no override configuration provided, controllers should display with their global defaults

## Actual behavior

Unless an override is configured, no controllers show.

## Your Setup (please complete the following information)

- Unity Version 2019.1.9
- XRTK Version 0.1.11

## Target Platform (please complete the following information)

- ALL platforms with controllers

SimonDarksideJ:
Resolved by #234

PR #234, opened 2019-07-08
Visualisation fix for #233

SimonDarksideJ:

## Overview

Initial check-in of Visualisation fix for #233


SimonDarksideJ:
In testing, found an issue with assigning models.

A) The Unity API must have updated as the previous Prefab check was failing (model could not be assigned) - Fixed
B) Even with the model assigned, the override model was not used - investigating
C) Walking the code, if a specific controller has specified to use the platform default, this is not being honoured - investigating

SimonDarksideJ:
All issues resolved, including the OculusTouch pointer issue which was mapped incorrectly to an axis instead of a button

XRTK-Core/Packages/com.xrtk.core/Definitions/Controllers/MixedRealityControllerVisualizationProfile.cs
```diff
@@ -99,20 +118,21 @@ public GameObject GlobalRightHandModel
         /// </summary>
         /// <param name="controllerType">The type of controller to query for</param>
         /// <param name="hand">The specific hand assigned to the controller</param>
-        public GameObject GetControllerModelOverride(Type controllerType, Handedness hand)
+        public bool GetControllerModelOverride(Type controllerType, Handedness hand, out GameObject overrideModel)
```
wassx: Minor:  missing param doc, maybe worth to mention that it will be set to null when no type was found

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseController.cs
```diff
@@ -244,21 +240,36 @@ internal async Task TryRenderControllerModelAsync(Type controllerType, byte[] gl
...
+            // If we've got a controller model, then place it in the scene and get/attach the visualizer.
             if (controllerModel != null)
             {
-                var controllerObject = UnityEngine.Object.Instantiate(controllerModel, MixedRealityToolkit.Instance.MixedRealityPlayspace);
-                controllerObject.name = $"{controllerType.Name}_{controllerObject.name}";
-                Visualizer = controllerObject.GetComponent<IMixedRealityControllerVisualizer>();
+                //If the model was loaded from a system template
+                if (useSystemDefaultModels && gltfObject != null)
+                {
+                    controllerModel.name = $"{controllerType.Name}_Visualization";
```
wassx: Could it be that controllerType gets null in here?

SimonDarksideJ: It's expected all the way through this function that you might never get a model, so yes, the end result could be null.  Hence all the fallbacks to an eventual catch-all "no model" condition

XRTK-Core/Packages/com.xrtk.core/Definitions/Controllers/MixedRealityControllerVisualizationProfile.cs
```diff
@@ -71,20 +71,39 @@ public GameObject GlobalLeftHandModel
...
             get => globalRightHandModel;
             private set => globalRightHandModel = value;
         }
+        
+        [SerializeField]
+        [Tooltip("Default Pose Action.\nNote: If an override profile is not found for a specific controller, the default pose action is used.")]
+        private MixedRealityInputAction globalPointerPose;
+
+        /// <summary>
+        /// The Default controller model, for when there is no specific controller model for the Left hand or when no hand is specified (Handedness = none)
```
Jarodshow: Is this the right summary?

PR #256, opened 2019-07-25
Updated Build Tools

StephenHodgson:

## Overview

Updated the Lumin build tools to include a way to install the package after the build completes successfully.

## Target of the change:

Is this enhancement for:

- Build Utilities

## Changes:

- added a way to automatically install the lumin package on successful build completion.
- cleaned up the build window a bit.


XRTK-Core/Packages/com.xrtk.core/Utilities/BuildAndDeploy/LuminPlayerBuildTools.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        /// <param name="buildInfo"></param>
+        public static BuildReport BuildPlayer(BuildInfo buildInfo)
+        {
+            if (!Application.isBatchMode)
+            {
+                buildInfo.PostBuildAction += PostBuildAction;
+            }
+
+            return UnityPlayerBuildTools.BuildUnityPlayer(buildInfo);
```
SimonDarksideJ: Would this not be better implemented as an extension class, calling base.UnityPlayerBuildTools, rather than this complex interweving.  

Lumin should be an override of the base rather than separate "bits"

XRTK-Core/Packages/com.xrtk.core/Utilities/BuildAndDeploy/BuildDeployWindow.cs
```diff
@@ -1007,25 +1006,49 @@ private async void PairDevice()
...
-            appxCancellationTokenSource = null;
+            buildCancellationTokenSource = new CancellationTokenSource();
+
+            switch (EditorUserBuildSettings.activeBuildTarget)
+            {
+                case BuildTarget.WSAPlayer:
+                    await UwpPlayerBuildTools.BuildPlayer(BuildDeployPreferences.BuildDirectory, cancellationToken: buildCancellationTokenSource.Token);
+                    break;
+                case BuildTarget.Lumin:
+                    LuminPlayerBuildTools.BuildPlayer(new BuildInfo());
```
SimonDarksideJ: I would have expected this call to be an override of the base BuildUnityPlayer call instead of hoping out and back again.

XRTK-Core/Packages/com.xrtk.core/Utilities/BuildAndDeploy/LuminPlayerBuildTools.cs
StephenHodgson: Agreed, or maybe needs an interface for the build tools that then calls the right method for the right build platform

XRTK-Core/Packages/com.xrtk.core/Utilities/BuildAndDeploy/LuminPlayerBuildTools.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        private static async void PostBuildAction(IBuildInfo buildInfo, BuildReport buildReport)
+        {
+            await new Process().RunAsync($@"/C mldb install -u ""{buildInfo.OutputDirectory}""", "cmd.exe", true);
```
StephenHodgson: This needs to be updated to also support mac/linux install.

XRTK-Core/Packages/com.xrtk.core/Utilities/BuildAndDeploy/LuminPlayerBuildTools.cs
StephenHodgson: Looked into this more. Because the `UnityPlayerBuildTools` class is static and is getting methods called directly by the editor, we cannot use derived classes to correctly call the appropriate overriding method.

StephenHodgson:
Ready for review

PR #261, opened 2019-08-05
Added CardinalAxis and FlattenMode enum definitions

StephenHodgson:

## Overview

Added additional enum definitions for cardinal axis and flattening mode.

## Target of the change:

Is this enhancement for:

## Changes:

- To be used with BoundingBox but possibly other uses.


SimonDarksideJ:
OK, adding two new enums, but they aren't being consumed anywhere, or are these changes needed for updates in the SDK?

PR #262, opened 2019-08-06
Dev/defininitions

StephenHodgson:
Fixed cardinal axis offset. (should be power of 2 ie: 1,2,4,8,16...)


SimonDarksideJ:
That was a quick review @StephenHodgson :P
But obviously ok :D

PR #266, opened 2019-08-06
Retiring of original wireframe shader due to compatibility issues

SimonDarksideJ:

## Overview

Removing current wireframe shader till we have a working cross-platform solution

## Target of the change:

Is this enhancement for:

## Changes:

- Removes wireframe shader


StephenHodgson:
I thought you were going to add the new one back in?

PR #271, opened 2019-08-07
Updated Wireframe shader with Android support

SimonDarksideJ:

## Overview

Updated wireframe shader to be Android compatible.  Still can't use it but it no longer causes errors.

## Target of the change:

Is this enhancement for:

## Changes:

- Fixes: XRTK_Wireframe.shader.


SimonDarksideJ:
Formatting updated

StephenHodgson:
Does this mean we can close #266?

SimonDarksideJ:
Updated shader to use a graphics tier configuration setting that is enabled on Lumin but disabled on Android.  Have a see if that helps.

PR #272, opened 2019-08-07
Fix/common root nre

StephenHodgson:

## Overview

Fixed an issue where finding the common root object could result in a null reference exception

## Target of the change:

Is this enhancement for:

- Extension

## Changes:

Checked if transform 1 is null and break the loop


XRTK-Core/Packages/com.xrtk.core/Extensions/TransformExtensions.cs
```diff
@@ -324,5 +329,21 @@ public static void SetCollidersActive(this Transform transform, bool isActive)
                 colliders[i].enabled = isActive;
             }
         }
+
+        /// <summary>
+        /// Sets the physics layer on this and all child <see cref="Transform"/>s with the provided value.
+        /// </summary>
+        /// <param name="transform"></param>
+        /// <param name="layer"></param>
+        public static void SetLayerRecursively(this Transform transform, int layer)
```
StephenHodgson: I thought we merged this change already?

PR #281, opened 2019-08-09
dev/spatial-awareness-refactor

StephenHodgson:

## Overview

Moved mesh display options from the data provider profile to the system profile for easier ability to track spatial mesh visibility.

## Target of the change:

Is this enhancement for:

## Breaking Changes:

- Marked `IMixedRealitySpatialAwarenessSystem.SetMeshVisibility` obsolete
- Marked `BaseMixedRealitySpatialMeshObserverProfile.MeshDisplayOption` obsolete
- Added `IMixedRealitySpatialAwarenessSystem.SpatialMeshVisibility`


XRTK-Core/Packages/com.xrtk.core/Providers/SpatialObservers/BaseMixedRealitySpatialMeshObserver.cs
```diff
@@ -31,7 +31,7 @@ protected BaseMixedRealitySpatialMeshObserver(string name, uint priority, BaseMi
             MeshLevelOfDetail = profile.MeshLevelOfDetail;
             MeshTrianglesPerCubicMeter = profile.MeshTrianglesPerCubicMeter;
             MeshRecalculateNormals = profile.MeshRecalculateNormals;
-            meshDisplayOption = profile.MeshDisplayOption;
+            meshDisplayOption = MixedRealityToolkit.SpatialAwarenessSystem.SpatialMeshVisibility;
```
SimonDarksideJ: This part seems odd, should this still be here?

StephenHodgson: Yes or else the initial mesh visibility options won't be set.

SimonDarksideJ: Fine , but if its part of the initialization of this class, why separate it out?
Seems odd as we always group settings together for convenience

StephenHodgson: We can probably move a new more settings up to the spatial awareness profile, but the main idea is that each device will have different performance requirements and therefore different settings. The visibility options was one setting that could be set in an agnostic way.

Not sure the rest are like this. Suggestions?

XRTK-Core/Packages/com.xrtk.core/Interfaces/SpatialAwareness/IMixedRealitySpatialAwarenessSystem.cs
```diff
@@ -81,6 +90,7 @@ public interface IMixedRealitySpatialAwarenessSystem : IMixedRealityEventSystem
         /// Sets the provided mesh <see cref="SpatialMeshDisplayOptions"/> on all the <see cref="DetectedSpatialObservers"/>
         /// </summary>
         /// <param name="displayOptions"></param>
+        [Obsolete("Use SpatialMeshVisibility instead")]
```
FejZa: Not sure whether using Obsolete / having obsolete APIs in a v0.1.5 Software that isn't bascially even released makes sense? Why not just get rid of it right away. What do you think?

StephenHodgson: I'm fine with this I guess since I made it pretty clear in the Pull Request overview.

XRTK-Core/Packages/com.xrtk.core/Providers/SpatialObservers/BaseMixedRealitySpatialMeshObserver.cs
SimonDarksideJ: My last comment didn't make it through.  Basically, (in reference to FR #290) If platform specific config is needed, it should be handled through a branch of config for that platform.  Else we descend in to madness having variants everywhere.
I would prefer we keep spatial settings together and if they are different between one platform and the next, then that is another profile and we need a better platform management system, rather than twist config.  make sense?

StephenHodgson: Nope, doesn't make much sense to me. 

I don't really like the proposal in #290 either because we only need one master configuration profile. The platform specific profiles should exist inside each system as part of the data provider configuration options (Which is platform specific)

StephenHodgson: In this special case, we wanna at least initialize the settings to a global one (which is purely just the visibility of the spatial meshes) but then give the user the option to override the visibility option if they so choose.

StephenHodgson: Can we revisit this later?

StephenHodgson: https://github.com/XRTK/XRTK-Core/issues/294

PR #282, opened 2019-08-09
dev/focus-cleanup

StephenHodgson:

## Overview

Cleaned up focus result a bit. We were inappropriately checking if it was null (should never be null)

## Target of the change:

Is this enhancement for:

## Changes:

- Removed a flag to set the pointer's `IsTargetPositionLockedOnFocusLock` to true if our `hitObject` is not the current pointer target.
- a bit of formatting
- Removed null checks for pointer.Result. This value should never be null.

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Services/InputSystem/FocusProvider.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Services/InputSystem/FocusProvider.cs
@@ -190,14 +190,6 @@ namespace XRTK.Services.InputSystem
 
             public void UpdateHit(PointerHitResult hitResult)
             {
-                if (hitResult.HitObject != CurrentPointerTarget)
-                {
-                    // Pointer.OnPreCurrentPointerTargetChange();
-
-                    // Set to default:
-                    Pointer.IsTargetPositionLockedOnFocusLock = true;
-                }
-
                 PreviousPointerTarget = CurrentPointerTarget;
 
                 focusDetails.Object = hitResult.HitObject;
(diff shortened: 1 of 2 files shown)


StephenHodgson:
@SimonDarksideJ not as confident in these changes, so I'd like a good review and test before we merge these

XRTK-Core/Packages/com.xrtk.core/Services/InputSystem/FocusProvider.cs
```diff
@@ -190,14 +190,6 @@ public PointerData(IMixedRealityPointer pointer)
 
             public void UpdateHit(PointerHitResult hitResult)
             {
-                if (hitResult.HitObject != CurrentPointerTarget)
-                {
-                    // Pointer.OnPreCurrentPointerTargetChange();
-
-                    // Set to default:
-                    Pointer.IsTargetPositionLockedOnFocusLock = true;
```
SimonDarksideJ: So what sets this to true now, locking it to a specific target?

StephenHodgson: In the inspectors or the developer toggles this flag in one of their scrips

StephenHodgson: This didn't look like it was needed. I'm not entirely sure what it was there to begin with

SimonDarksideJ: If we are unsure, then we should poll the community and even query it on Mrtk as to its use case and validity. As we do with any original code

StephenHodgson: I think you missed the point.

This doesn't actually do anything afaik. I'm not interested in how the MRTK does it, or its usage there as it's likely been changed.

StephenHodgson:
Gonna revisit this

PR #286, opened 2019-08-13
Dev/pointer refactor

StephenHodgson:

## Overview

Refactoring pointers to better facilitate object manipulation mechanics. 

The `IMixedRealityPointer.Result` can never be null, and added an Offset calculation from the hit result and pointer position has been added to understand the pivot location for that manipulation event.

## Target of the change:

Is this enhancement for:


XRTK-Core/Packages/com.xrtk.core/Interfaces/InputSystem/IMixedRealityCursor.cs
```diff
@@ -17,15 +17,10 @@ public interface IMixedRealityCursor : IMixedRealityFocusChangedHandler, IMixedR
...
         /// <summary>
-        /// Surface distance to place the cursor off of the surface at
+        /// Surface distance to place the cursor off of the surface at the pointers raycast hit position.
         /// </summary>
         float SurfaceCursorDistance { get; }
 
-        /// <summary>
-        /// The maximum distance the cursor can be with nothing hit
-        /// </summary>
-        float DefaultCursorDistance { get; set; }
```
StephenHodgson: This is always the pointer's current extent

PR #287, opened 2019-08-13
Release 0.1.16

StephenHodgson:
The controller namespace adjustment requires us to make a new point release for future SDK work.


SimonDarksideJ:
Requires #288

PR #291, opened 2019-08-15
dev/pointer-refactor-2

StephenHodgson:

## Overview

Closed the old Pointer Refactor to try and make smaller, more direct changes.

## Breaking Changes

- Moved all the focus data up a level
  - All references to `IMixedRealityPoniter.Result.Details` can now be found in the `Result`
  - Removed all null checks for `Result` as it can never be null.

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Services/InputSystem/GenericPointer.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Services/InputSystem/GenericPointer.cs
@@ -72,6 +72,7 @@ namespace XRTK.Services.InputSystem.Pointers
         /// <inheritdoc />
         public bool IsFocusLocked { get; set; }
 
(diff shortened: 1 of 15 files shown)


XRTK-Core/Packages/com.xrtk.core/Services/InputSystem/GenericPointer.cs
```diff
@@ -72,6 +72,8 @@ public virtual IMixedRealityInputSource InputSourceParent
         /// <inheritdoc />
         public bool IsFocusLocked { get; set; }
 
+        public bool SyncPointerTargetPosition { get; set; }
```
StephenHodgson:
```suggestion
        /// <inheritdoc />
        public bool SyncPointerTargetPosition { get; set; }
```

PR #305, opened 2019-08-31
Fixed a warning during first start

StephenHodgson:

## Overview

Fixed a warning when forcing the asset database to refresh during reload. This may or may not fix the first time startup crash we're seeing but worth a test.

## Target of the change:

Is this enhancement for:

## Changes:

- Fixes: Don't force refresh of asset database during startup, per warning from Unity Editor.


SimonDarksideJ:
Doesn't fix the crash but doesn't cause any other errors

PR #307, opened 2019-09-03
Fix to resolve Git Tag sorting issue #300

### LINKED ISSUE #300: Latest package versions don't properly get pulled down.
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

If the upm package extension doesn't get installed first, then the later packages will fail to install the latest versions.

## To Reproduce

1. Create a new project
2. Import the seed package or manually reference the xrtk package in the manifest.json
3. Wait for import

## Your Setup (please complete the following information)

- Unity Version 2019.1.14f1
- XRTK Version 0.1.17
- XRTK.SDK 0.1.12
- XRTK.UPM 0.9.6

## Additional context

Seems to be happening because the UPM extension package isn't installed first?

In either case, the versioning is not putting the 0.1.12 ahead of 0.1.9 (which is the version that installed)

SimonDarksideJ:

## Overview

fix to resolve #300, properly sorts the tag versions returned by Git based on a 3 digit version code.  Select the actual latest version based on:

x.x.x version number

Previously, last was determined by the double value parsed value, which is incorrectly sorted by .NET.  Now sorts by each individual version identifier to locate the latest.
## Target of the change:

Is this enhancement for:

## Changes:

Updates Git Tag version sort to sort by each individual version identifier

- Fixes: #300

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Inspectors/Utilities/Packages/MixedRealityPackageUtilities.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Inspectors/Utilities/Packages/MixedRealityPackageUtilities.cs
@@ -209,6 +209,7 @@ namespace XRTK.Inspectors.Utilities.Packages
 
         private static async Task AddPackageAsync(MixedRealityPackageInfo packageInfo)
         {
+            var versionSeperator = new char[] { '.' };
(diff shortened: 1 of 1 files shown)


XRTK-Core/Packages/com.xrtk.core/Inspectors/Utilities/Packages/MixedRealityPackageUtilities.cs
```diff
@@ -209,7 +209,8 @@ internal static async void ValidatePackages()
 
         private static async Task AddPackageAsync(MixedRealityPackageInfo packageInfo)
         {
-            var tag = (await GitUtilities.GetAllTagsFromRemoteAsync(packageInfo.Uri)).LastOrDefault();
+            var versionSeperator = new char[] { '.' };
```
StephenHodgson: This could be a readonly field in the class so we don't allocate it every time we add a package.

PR #308, opened 2019-09-04
Oculus Controller mapping update

SimonDarksideJ:

## Overview

Resolving mapping issue originally identified in MRTK where near touch data was not identified and the "Pressed" call was called incorrectly.

Now in the OpenVR / Generic Joystick spec, digital buttons whether as a button or axis are ratified in the Button code, single-axis only axis data is handled separately.

## Target of the change:

Is this enhancement for:

- Platform (new or updated platform)

## Changes:

- updates GenericJoystickController to identify buttons, inc axis buttons, in button code and single-axis only axis data separately
- Update Oculus touch mapping to get Trigger pressed from the actual trigger data (not near touch)

## Dependancies:

SDK change required to update default profiles.


XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/UnityInput/GenericJoystickController.cs
```diff
@@ -85,10 +85,24 @@ protected void UpdateButtonData(MixedRealityInteractionMapping interactionMappin
         {
             Debug.Assert(interactionMapping.AxisType == AxisType.Digital);
 
-            var keyButton = Input.GetKey(interactionMapping.KeyCode);
-
             // Update the interaction data source
-            interactionMapping.BoolData = keyButton;
+            switch (interactionMapping.InputType)
+            {
+                case DeviceInputType.TriggerPress:
```
wassx: So the now removed: `case DeviceInputType.ButtonPress` was wrong? This should now handled in the `default:` ?

SimonDarksideJ: Yes, button press and and other digital inputs that are not axis are handled in the default

But this does pose another case. This is based on unitys openvr spec. Maybe we need an additional bit of meta data about the input to identify source axis and target axis to better identify how to transform the data. Doing it based on the input type is potentially too limiting.

Will stick to how this is now as this solves the immediate issue. 

But there has been an additional comment in the MRTK space I want to review before merging, about getaxis and getrawaxis

SimonDarksideJ:
Lol, I'd put this in progress to prevent merging as there was a debate on GetAxis vs GetAxisRaw on MRTK.  But if changes are needed I'll just do another PR :S

StephenHodgson:
Sorry mate

SimonDarksideJ:
We simply need to be sure to check flags when set, especially if it's "Ready for Review" :D

StephenHodgson:
I think there's github hooks to make sure those labels are there before we can merge. But not sure

PR #313, opened 2019-09-10
Added checks for editor updating

StephenHodgson:

## Overview

Added more checks to see if the editor is already attempting to update to try and mitigate a warning seen during initial import.

## Target of the change:

Is this enhancement for:


SimonDarksideJ:
Looks ok, but will test in the morning to confirm

PR #326, opened 2019-09-15
Dev/editor platform flag

StephenHodgson:

## Overview

Added editor platform flag to the supported platform enum to specify editor only services

## Target of the change:

Is this enhancement for:

## Breaking Changes:

- Breaks: All existing data provider profiles. All profiles will need to add the `Editor` flag to run in the editor by default, and update their platforms.

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseController.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseController.cs
@@ -213,11 +213,10 @@ namespace XRTK.Providers.Controllers
 
             if (!visualizationProfile.RenderMotionControllers) { return; }
 
-            GameObject controllerModel = null;
             GltfObject gltfObject = null;
 
             // If a specific controller template exists, check if it wants to override the global model, or use the system default specifically (in case global default is not used)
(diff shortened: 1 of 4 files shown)


XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseController.cs
```diff
@@ -213,11 +213,10 @@ internal async Task TryRenderControllerModelAsync(Type controllerType, byte[] gl
 
             if (!visualizationProfile.RenderMotionControllers) { return; }
 
-            GameObject controllerModel = null;
             GltfObject gltfObject = null;
 
             // If a specific controller template exists, check if it wants to override the global model, or use the system default specifically (in case global default is not used)
-            bool useSystemDefaultModels = visualizationProfile.GetControllerModelOverride(controllerType, ControllerHandedness, out controllerModel);
+            bool useSystemDefaultModels = visualizationProfile.GetControllerModelOverride(controllerType, ControllerHandedness, out var controllerModel);
```
FejZa: Not a fan of var here but I guess that's personal preference.

StephenHodgson: Just a product of in-lining the out variable. I don't really mind, more concise.

Issue #325, opened 2019-09-14
XRTK-Core 0.1.19 does not  pull all dependencies


milos12345:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

Edited manifest as per instructions for XRTK-Core 0.1.19

## To Reproduce

New Scene, Edit manifest.json per instructions in Release

## Expected behavior

All dependencies are downloaded (it worked for 0.1.18)

## Actual behavior

Only XRTK.Core is downloaded

## Your Setup (please complete the following information)

- Unity Version [e.g. 2019.1.14f1]
- XRTK Version [e.g. 0.1.19]

## Target Platform (please complete the following information)

- WMR immersive
- OpenVR

StephenHodgson:
What dependencies were not pulled as expected?

StephenHodgson:
We removed it because it was causing crashing errors and longer startup times. 

Once the core package is in the project you should be able to add all of the packages via upm package manager

milos12345:
Got it, I prefer it the new way. It will be helpful to make a note in Release section that it should be done through Package Manager. In 5 years working with Unity I have not needed the Package Manager, so I would not even look there, and there are probably more users like me.

StephenHodgson:
Will do! thanks for the feedback 😃

StephenHodgson:
Looks like we'll attempt to at least pull in the SDK and let users decide which platforms they want to install.

See #333 333

PR #333, opened 2019-09-16
pull the sdk and core by default

StephenHodgson:

## Overview

Pulls the core and sdk by default

- Addresses #325


XRTK-Core/Assets/XRTK.Seed/MixedRealityPackageSeed.cs
```diff
@@ -66,7 +66,7 @@ private static void Run()
 
                         File.WriteAllText(manifestFilePath, text);
 
-                        Client.Add("com.xrtk.core");
+                        Client.Add("com.xrtk.sdk");
```
FejZa: Assuming core gets added as well since SDK has a dependancy on it?

StephenHodgson: Correct

Issue #259, opened 2019-08-03
Consider implementing a "TryGet" pattern for Services/Providers


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Feature Request

## Is your feature request related to a problem? Please describe

A standard pattern when querying a registry for entries where the return may be null, is to use a "TryGet" approach, returning true / false based on whether the value was returned.

This would reduce implementation code for checking if the service was returned or not.

## How would you classify your suggestion

- Usability / Configuration
- Architecture / Services

## Describe the solution you'd like

Any service lookup should also implement a TryGet method in line with .NET standards

## Additional context

MRTK recently refactored their service registry, implementing TryGet methods, which seems valid
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5507/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.md#diff-49dd688146666f63e319c5d422a7ec10

PR #334, opened 2019-09-16
Update to enable TryGetService calls for the MixedRealityToolkit

SimonDarksideJ:

## Overview

Update to enable TryGetService calls for the MixedRealityToolkit

## Target of the change:

Is this enhancement for:

## Changes:

* Adds TryGetService
* Adds TryGetService (By Name)
* Includes Test fixtures to validate new API
- Closes: #259


StephenHodgson:
Do you have any suggestions where we should replace some of the calls for this?

Issue #345, opened 2019-09-18
For controls that use a single axis, they are creating their config asset incorrectly


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

Post the native Oculus update, when a new config asset is created that has single axis configuration, it is incorrectly being identified as "custom" data (as used by Oculus today)

## To Reproduce

1. Create new OpenVR Controller profile
2. Examine asset to find "InputName" populated instead of "AxisX" for single axis config

## Expected behavior

Single Axis configuration should be recorded in the AxisX field

## Actual behavior

Single Axis configuration is being recorded in the custom "InputName" field

## Your Setup (please complete the following information)

- Unity Version 2019.1.14
- XRTK Version 0.1.19

## Target Platform (please complete the following information)

- WMR immersive
- OpenVR

## Additional context

Simply requires single axis configuration to provide an empty second axis

StephenHodgson:
So the constructor is malformed?

SimonDarksideJ:
No, the constructor is fine.  However with the requirement to have custom data for Oculus inputs, this overrides the "defaulted" approach for Axis data.  So inputs requiring axis data need to provide two axis.

I support I could also update the Axis based constructor to make the second argument not optional.

StephenHodgson:
The constructor is definitely incorrect. Let's fix it.

StephenHodgson:
Goes back to https://github.com/XRTK/XRTK-Core/pull/237 where we introduced the `inputName` constructors.

The constructors are ambiguous to the existing ones that have default parameters, so it's getting confused with `axisCodeX` instead of `inputName`

StephenHodgson:
We need to update the new constructors by swapping the `inputName` with the `inputAxisType` this will help ensure we don't have ambiguity.

Issue #336, opened 2019-09-17
Error Input Axis  is not setup.


milos12345:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

On Play, this error is logged every frame

```
Input Axis  is not setup.
 To change the input settings use: Edit -> Settings -> Input
  at (wrapper managed-to-native) UnityEngine.Input.GetAxisRaw(string)
  at XRTK.Providers.Controllers.UnityInput.GenericJoystickController.UpdateButtonData (XRTK.Definitions.Devices.MixedRealityInteractionMapping interactionMapping) [0x00052] in D:\M\Template\XRTKTemplate9\Library\PackageCache\com.xrtk.core@0.1.19\Providers\Controllers\UnityInput\GenericJoystickController.cs:92 
  at XRTK.Providers.Controllers.UnityInput.GenericJoystickController.UpdateController () [0x00086] in D:\M\Template\XRTKTemplate9\Library\PackageCache\com.xrtk.core@0.1.19\Providers\Controllers\UnityInput\GenericJoystickController.cs:57 
  at XRTK.Providers.Controllers.OpenVR.GenericOpenVRController.UpdateController () [0x0007e] in D:\M\Template\XRTKTemplate9\Library\PackageCache\com.xrtk.core@0.1.19\Providers\Controllers\OpenVR\GenericOpenVRController.cs:164 
  at XRTK.Providers.Controllers.UnityInput.UnityJoystickDataProvider.Update () [0x00059] in D:\M\Template\XRTKTemplate9\Library\PackageCache\com.xrtk.core@0.1.19\Providers\Controllers\UnityInput\UnityJoystickDataProvider.cs:54 
  at XRTK.Services.MixedRealityToolkit.UpdateAllServices () [0x00095] in D:\M\Template\XRTKTemplate9\Library\PackageCache\com.xrtk.core@0.1.19\Services\MixedRealityToolkit.cs:1284 
UnityEngine.Debug:LogError(Object)
XRTK.Services.MixedRealityToolkit:UpdateAllServices() (at Library/PackageCache/com.xrtk.core@0.1.19/Services/MixedRealityToolkit.cs:1288)
XRTK.Services.MixedRealityToolkit:Update() (at Library/PackageCache/com.xrtk.core@0.1.19/Services/MixedRealityToolkit.cs:705)
```

## To Reproduce

Open Unity Hub
New Project with 2019.1.14f1
Choose 3D
[...]

PR #346, opened 2019-09-18
Resolves issue where Single Axis data is incorrectly assigned in new profiles

SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Change Request

## Overview

Resolves issue where Single Axis data is incorrectly assigned in new profiles
Also adds extra debugging information for Unity Axis failures
resolves #345

Required for #344

## Target of the change:

Is this enhancement for:

- Core (core framework, interfaces and definitions)

## Changes:

Brief list of the targeted features that are being changed.

- Fixes: #345

PR #347, opened 2019-09-18
Fixed the ambiguous interaction mapping constructors

StephenHodgson:

## Overview

Fixed the ambiguous interaction mapping constructors.

## Target of the change:

Is this enhancement for:

## Changes:

- Fixes: #345 
- Fixes: #336
- Supersedes: #346 

## Breaking Changes:

- Update the `MixedRealityInteractonMapping` constructors but afaik the Oculus package is the only one taking advantage of it atm.

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Definitions/Devices/MixedRealityInteractionMapping.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Definitions/Devices/MixedRealityInteractionMapping.cs
@@ -91,7 +91,7 @@ namespace XRTK.Definitions.Devices
         /// <param name="axisType">The axis that the mapping operates on, also denotes the data type for the mapping</param>
         /// <param name="inputType">The physical input device / control</param>
         /// <param name="inputName">Optional inputName value to get input for a coded input identity from a provider</param>
-        public MixedRealityInteractionMapping(uint id, string description, AxisType axisType, DeviceInputType inputType, string inputName)
+        public MixedRealityInteractionMapping(uint id, string description, AxisType axisType, string inputName, DeviceInputType inputType)
         {
             this.id = id;
             this.description = description;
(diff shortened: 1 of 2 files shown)


SimonDarksideJ:
This addresses only the changes for oculus, however, it also breaks all existing mappings for all other controllers.  This needs to be based off this fix branch.
Also, it will need the default assets updating with this fix.  Let's close this PR and I'll incorporate your changes in my branch.

StephenHodgson:
>  however, it also breaks all existing mappings for all other controllers. This needs to be based off this fix branch.

I don't understand. This shouldn't break any existing mappings.

> Let's close this PR and I'll incorporate your changes in my branch.

I don't think that's a good idea.

XRTK-Core/Packages/com.xrtk.core/Definitions/Devices/MixedRealityInteractionMapping.cs
```diff
@@ -189,6 +189,10 @@ public MixedRealityInteractionMapping(uint id, string description, AxisType axis
             keyCode = KeyCode.None;
         }
 
+        /// <summary>
+        /// Creates a copy of a <see cref="MixedRealityGestureMapping"/>
```
StephenHodgson:
```suggestion
        /// Creates a copy of a <see cref="MixedRealityInteractionMapping"/>
```

StephenHodgson:
> Not a fan of mixing up the order of parameters like this, but it works.

Any suggestions on how this could be better?

Mainly the issue is that the wrong Constructor is being called because of the order of the parameters.
One possible fix is to make the type different, but I'm not sure what would be an appropreate change.

The only other possibility is removing the `inputName` altogether and using the established `inputAxisX` and `inputAxisY` values that essentially function the same.

SimonDarksideJ:
> The only other possibility is removing the inputName altogether and using the established inputAxisX and inputAxisY values that essentially function for the same purpose.

No, they are defined concepts and should not be reused.

It's an acceptable fix

PR #358, opened 2019-09-25
First pass at input action refactor

StephenHodgson:
Preliminary work on Input Action Profile refactor.

I'll try to list as many of the changes as possible going into this. 

- Simplified the controller implementations a bit by providing a `InteractionMapping` wrapper for raising all the events to the input system.
- Removed all the rule profiles and data structures.
- Stubbed in the input action profile, but there's still a lot that needs to go in there.


XRTK-Core/Packages/com.xrtk.core/package.json
```diff
@@ -10,7 +10,7 @@
     "Mixed Reality",
     "DI"
   ],
-  "version": "0.1.20",
+  "version": "0.1.21",
```
SimonDarksideJ: A little premature updating this for a feature branch :D

StephenHodgson: Nope. Always increment for breaking changes.

SimonDarksideJ: Doctors also used to always just saw legs off at the first sign of infection.   Never say "Always"

PR #360, opened 2019-09-27
Feature - Native Services

StephenHodgson:

## Overview

Adds a way to create services for native data providers. This enables the ability to hot reload native libraries in the editor as well as providing an easy way to access native libraries without having to worry about the interop between the native and managed api.

## Target of the change:

Is this enhancement for:

- Extension

## Changes:

- Provides developers an easy way to define and use native libraries as data providers.


SimonDarksideJ:
One question on this approach, it seems to separate Native Libraries as new providers, making another data provider sku.
Is this what is intended, or are these dll's that are registered that can be called from a Data provider implementation? @StephenHodgson

StephenHodgson:
> Is this what is intended, or are these dll's that are registered that can be called from a Data provider implementation?

Yes, each native library can be used as a service data provider, with a platform agnostic service wrapper.

PR #364, opened 2019-09-29
First pass on input action refactor

StephenHodgson:
Preliminary work on Input Action Profile refactor.

I'll try to list as many of the changes as possible going into this. 

- Simplified the controller implementations a bit by providing a `InteractionMapping` wrapper for raising all the events to the input system.
- Removed all the rule profiles and data structures.
- Stubbed in the input action profile, but there's still a lot that needs to go in there.


StephenHodgson:
@SimonDarksideJ I think this is ready to go into the feature branch as a list of initial changes.

Thoughts?

SimonDarksideJ:
I'm fine with it going to a Feature branch as a started, but obviously it only removed rules and doesn't actually "do" anything yet as far as I can see.

PR #374, opened 2019-10-01
Ensure we always start a service if only the editor flag is set

StephenHodgson:

## Overview

There was one use case for the `PlatformUtility` flags that wasn't covered. Where we only want a service to run in the editor only on any platform.

| Figure | Editor | Rumtime |
|---|---|---|
| 1 | X | |
| 2 | X | X |
| 3 | | X |

1. This should **always** run in the editor regardless of the target build platform but never the build:image

2. This should run in the editor && build if the build target **IS** enabled:image

3. This should **never** run in the editor and only runs on the build:image

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Utilities/PlatformUtility.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Utilities/PlatformUtility.cs
@@ -85,16 +85,19 @@ namespace XRTK.Utilities
         {
             var isEditorSupported = IsEditorSupported(platforms);
 
-            if ((int)platforms == -1)
+            // If any platform set or if only the editor flag is set
+            if ((int)platforms == -1 || platforms == SupportedPlatforms.Editor)
             {
                 return true;
             }
 
-            if (!isEditorSupported || platforms == 0)
+            // If no platform set or if the editor is not supported
+            if (platforms == 0 || !isEditorSupported)
             {
                 return false;
             }
 
+            // If editor is supported and the current build target is enabled
             var target = GetSupportedPlatformMask(editorBuildTarget);
             return IsPlatformSupported(target, platforms);
         }


FejZa:
Will editor checked and lumin checked run when in lumin but not Editor runtime? Aka when deployed.

FejZa:
I would test it myself but I am on mobile and reading the description this popped into my head.

StephenHodgson:
> Will editor checked and lumin checked run when in lumin but not Editor runtime? 

I can't wrap my head around what you are describing 😅

FejZa:
Well, sorry. I was asking whether it's lumin && editor or it is lumin || editor. Or "can I have a service that will run in editor but also run in the platform build?"

StephenHodgson:
Just updated with a table

StephenHodgson:
It's `platform && editor`

FejZa:
But from your updated description I read the answer is yes

StephenHodgson:
I wasn't sure how to address `platform || editor` but I suspect that can be done in the implementation of the service itself.

FejZa:
Looks good to me. Can we get this merged? I need it for hand tracking to work properly in editor.

SimonDarksideJ:
I think this is a good start, although we do need to review the supported platform model.  However, I see 4 possibilities not 3 in combination with an editor:

* Editor Only - Service that only runs in the editor (mock) - any platform
* Editor AND Platform - Service that only runs in the editor (mock) - specific platform
* Platform(s) Only - Service that only runs on the platform (actual)
* Editor OR Platform(s) - Service that is designed to work both in the editor and on any platform

We'll just have to come up with the additional configuration patterns later

StephenHodgson:
> We'll just have to come up with the additional configuration patterns later

Would you like to write up a good task for this?

SimonDarksideJ:
I should, yes :D

PR #405, opened 2019-11-05
Fixed a few things with the new camera rig

### LINKED ISSUE #404: When stopping the project in the editor, an error is received about destroying the camera
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

Since the 0.1.25 update, now when the Editor stops playing, an error is generated about destroying the camera

```
Can't destroy Transform component of 'Main Camera'. If you want to destroy the game object, please call 'Destroy' on the game object instead. Destroying the transform component is not allowed.
```

## To Reproduce

1. Create a new XRTK
2. Configure the scene
3. Run project
4. Stop Editor
5. Observe error

## Expected behavior

No errors

## Actual behavior

Above error reported

## Your Setup (please complete the following information)

- Unity Version 2019.1.14
- XRTK Version 0.1.25

## Target Platform (please complete the following information)

- OpenVR

StephenHodgson:

## Overview

Fixed the errors from the camera rig shutting down improperly and syncing the playspace and bodypace transforms correctly.

Mainly we wanna ensure that the body is roughly the same position/rotation of the head (within the adjustment angle threshold)

## Target of the change:

Is this enhancement for:

## Changes:

- Fixed #404 component casting check to ensure we're not trying to destroy components that aren't a IMixedRealityCameraRig type.
- Added flags to check if app was quitting before setting up objects when we're trying to cleanup.
- Reverted some changes where we were parenting to the body instead of the playspace.
- Reverted changes where we were moving the PlaySpace improperly.
- Added appropriate fallbacks to the cashed camera reference if the camera system isn't enabled

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Services/CameraSystem/MixedRealityCameraSystem.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Services/CameraSystem/MixedRealityCameraSystem.cs
@@ -189,7 +189,8 @@ namespace XRTK.Services.CameraSystem
                 }
             }
 
-            if (CameraRig is Component component)
+            if (CameraRig is Component component &&
+                component is IMixedRealityCameraRig)
             {
                 if (Application.isPlaying)
                 {
--- a/XRTK-Core/Packages/com.xrtk.core/Services/CameraSystem/MixedRealityCameraSystem.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Services/CameraSystem/MixedRealityCameraSystem.cs
@@ -170,12 +170,14 @@ namespace XRTK.Services.CameraSystem
         {
             base.Disable();
 
-            if (CameraRig == null) { return; }
+            var camera = CameraCache.Main;
 
-            if (CameraRig.CameraTransform != null)
+            if (camera != null)
             {
-                CameraRig.CameraTransform.SetParent(null);
+                camera.transform.SetParent(null);
             }
+
+            if (CameraRig == null) { return; }
 
             if (CameraRig.PlayspaceTransform != null)
             {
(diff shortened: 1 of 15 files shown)


XRTK-Core/Packages/com.xrtk.core/Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -223,5 +223,18 @@ private void ApplySettingsForTransparentDisplay()
             CameraCache.Main.nearClipPlane = profile.NearClipPlaneTransparentDisplay;
             QualitySettings.SetQualityLevel(profile.TransparentQualityLevel, false);
         }
+
+        private void SyncRigTransforms()
```
StephenHodgson: @SimonDarksideJ this is what I'd like to review specifically.

StephenHodgson:
@SimonDarksideJ tested this on WMR but Open VR and Oculus still need to be tested.

SimonDarksideJ:
Sadly, utter fail, for all the same reasons we attached everything to the Body.  Altering the headpose, does not alter all the things you want to "synchonise" with the head, e.g. Controllers, Teleportation, gaze, etc.
So this update breaks everything that requires the headpose to be updated, hence why it works find on WMR and OpenVR, which automatically translates the pose in the camera.

StephenHodgson:
Gonna close this for now. Yeah we def shouldn't be changing the playspace transform.

We expect that anyone will move that to re-position their players. This would def break in that scenario

SimonDarksideJ:
SO before we can merge this, we need to address the height set by platform (or manually) issue. Else we are regressing what is currently in public

StephenHodgson:
Agreed.

SimonDarksideJ:
Oculus PR unblocked and tested working with this new setup. Good to go.

StephenHodgson:
I'm gonna hold off on merging this until the submodules get their release

StephenHodgson:
Should be ready now

Issue #378, opened 2019-10-02
Improve Platform support configuration to handle multiple variants correctly


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Feature Request

## Is your feature request related to a problem? Please describe

Current Platform evaluations for services do not cover the full extent of the potential requirements, given that services require the ability to operate in the following conditions:

1. Editor Only - Service that only runs in the editor (mock) - any platform
2. Editor AND Platform - Service that only runs in the editor (mock) - specific platform
3. Platform(s) Only - Service that only runs on the platform (actual)
4. Everything - Service that is designed to work both in the editor and on any platform

Currently, only the first three modes are supported based on #374

As we expand the list of potential platforms (#377) these capabilities need to be reviewed

## How would you classify your suggestion

- Architecture / Services

## Describe the solution you'd like

(Leading from the example given in #374)

| Figure | Editor | Runtime | Description |
|---|---|---|---|
| 1 | X | | Only Editor |
| 2 | X | X | Both Editor and Runtime |
| 3 | | X | Only Runtime |
|4|X|(or) X | Everything |

1. This should run **ONLY** in the editor regardless of the target build platform but never the build:![image](https://user-images.githubusercontent.com/13334553/65984542-ff65ff00-e44d-11e9-8871-a0c34ebbab82.png)

2. This should run **ONLY** in the editor if the build target **IS** Selected:![image](https://user-images.githubusercontent.com/13334553/65984558-08ef6700-e44e-11e9-836f-0b7895db785d.png)

3. This should **never** run in the editor and only runs on the build:![image](https://user-images.githubusercontent.com/13334553/65984700-4eac2f80-e44e-11e9-955b-bb311d15ef83.png)

4. This can be run in the Editor and any build target
(Image tbc)

## Describe alternatives you've considered

No alternatives currently exist

## Additional context

Related to #377

StephenHodgson:
We already do this sans case 4.

The 4th case could easily be handled in the service implementation.

SimonDarksideJ:
Agreed, this is simply setting the requirement for the entire ask (so nothing gets missed in testing)

WIth 4 however, given the platform Utility is currently dictating when a service can be run, you couldn't do that in the service implementation :D But it's something to review.  Maybe by having a service define it's "supportability"

StephenHodgson:
I am open to doing something similar to #377 though

StephenHodgson:
[...]

Issue #377, opened 2019-10-02
Improve Supported platform support to handle multiple SDK's on a single OS


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Feature Request

## Is your feature request related to a problem? Please describe

Currently, Unity (and XRTK) only support an Operating system-level platform selection, unless a specific exception is made (e.g. Lumin / Xbox). However, more and more vendors rather than integrating directly with Unity are simply making platforms available under existing OS models (e.g. Oculus on Windows and Android)

This is creating complications when configuring Services / SDK's to operate or customising options within the XRTK.

## How would you classify your suggestion

- Usability / Configuration
- Architecture / Services

## Describe the solution you'd like

The solution needs to be able to provide and support a richer set of Platform definitions covering all the options for running solutions, e.g.

* Existing Options (Windows / Android / UWP / Lumin / etc)
* Windows OpenVR
* Android OpenVR
* UWP OpenVR
* Windows Oculus
* Android Oculus Quest
* Android Oculus Go
* etc...
(Full list needs to be determined and evaluated, given more mobile Style platforms are choosing Android)

> A possible solution has been described in the MRTK project here:
> https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4235

Ultimately, each platform variation needs to be able to also define whether is it active / enabled with an API check (which may also require platform variations) 
e.g. 

> Example only for reference

```
public class AndroidPlatformSupport : IPlatformSupport
{
    public virtual bool Check()
   {
        return Application.platform == RuntimePlatform.Android;
   }
}

public class OculusGoPlatformSupport : AndroidPlatformSupport
{
    public override bool Check()
    {
        return base.Check() && OVRPlugin.GetSystemHeadsetType() == SystemHeadset.Oculus_Go;
    }
}
```

This will lead to an improved method of validating and only loading Providers and Services for specific platforms

## Describe alternatives you've considered

Current practices are insufficient

Issue #247, opened 2019-07-17
Controller management required for situations where multiple data providers can exist on a single platform


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

With Windows Mixed Reality, OpenVR was for standalone and WMR was for UWP, there was a clear separation

Now with the addition of Oculus, we have situations where multiple data providers can compete for the controllers to manage, resulting in multiple data providers capturing and rendering the detected controllers.

Seems we need a higher level controller management framework to ensure that once a controller is activated and registered, it is ignored by any other provider.

## To Reproduce

1. Enable Joystick / OpenVR & Oculus on Windows Standalone (default behaviour)
2. Run the project with Oculus VR as the primary VR target
3. Observe that multiple controllers exist in the scene, some tracked and some not

## Expected behavior

Once a controller is active, it should be ignored by other data providers

## Actual behavior

Free for all, everyone registers and renders

## Your Setup (please complete the following information)

- Unity Version 2019.1.10
- XRTK Version 0.1.13

## Target Platform (please complete the following information)

- Windows Standalone

SimonDarksideJ:
Key Discussion point here, is what should be used as the central management point between all data providers? 

* Is it something new 
* or attached to the XRTK Entry point (Dictionary on the XRTK entry point for example)
* Or attached to the input system (A function call to validate if an input source exists for a given controller)

StephenHodgson:
Not really a bug. It's a matter of implementation details I suppose.

WMR and OpenVR seem to work just fine.

StephenHodgson:
Isn't this just a duplicate of #240?

PR #416, opened 2019-12-16
Feature - Platform System

StephenHodgson:
Added required classes, profiles, and inspectors

## Overview

The platform system aims to solve the problems we're having with multiple build targets on a single hardware platform.

## Target of the change:

Is this enhancement for:

# Changes

- [x] Fixes: #378
- [x] Fixes: #377
- [x] Fixes: #247


StephenHodgson:
Looks like the Unit Tests didn't like this change. Gotta look into that later.

SimonDarksideJ:
Seems fair, but not clear on how a platform declares itself yet

StephenHodgson:
> Seems fair, but not clear on how a platform declares itself yet

Getting there, ideally I think we want each upm package to register itself automagically when it's included/imported

SimonDarksideJ:
That I agree with, however we do need at least one example to compare, I'd prefer it to be Android based but Standalone should be sufficient to test

StephenHodgson:
Okay from here I'm going to suggest that we use this PR as a way to view the running changes to this feature and open new PRs against it as we tackle different aspects

XRTK-Core/Packages/com.xrtk.core/Interfaces/Platform/IMixedRealityPlatform.cs
```diff
@@ -8,5 +8,9 @@ namespace XRTK.Interfaces.PlatformSystem
     /// </summary>
     public interface IMixedRealityPlatform : IMixedRealityDataProvider
     {
+        /// <summary>
+        /// Is this platform currently active and available?
+        /// </summary>
+        bool IsActive { get; }
```
SimonDarksideJ: Not sure about the "IsActive" approach, surely as platforms are "active" they will simply register their platform with the Toolkit.  By the Platform being registered it will be active.
Shouldn't need to test if a platform is "Not" active.

StephenHodgson: Fair enough, I was also thinking that too, but for the built in platforms there's really no way to distinguish them bc they're all always registered but not necessarily "On", make sense?

StephenHodgson: Also, I'm not keen on the idea of services self-registering themselves. In fact I'd rather guard against that sort of thing, unless it's a trusted module.

SimonDarksideJ:
Can we break out some of the non-pr related changes (adding regions, adding null checks, etc) as it's making this PR harder to read as to what the changes are.  Most are goo they just need separately merging in @StephenHodgson

XRTK-Core/Packages/com.xrtk.core/Services/PlatformSystem/Platforms/EditorPlatform.cs
```diff
@@ -24,16 +25,12 @@ public EditorPlatform(string name, uint priority)
...
-#endif
-            }
-        }
+        public override string Name => "Editor";
+
+        /// <inheritdoc />
+        public override uint Priority => 0;
+
+        /// <inheritdoc />
+        public bool IsActive => Application.isEditor;
```
SimonDarksideJ: This method is a lot better, but can we rename "IsActive" to "IsAvailable", as we aren't checking if it's active, rather that it exists to be used

XRTK-Core/Packages/com.xrtk.core/Interfaces/Platform/IMixedRealityPlatform.cs
SimonDarksideJ: As discussed, can we rename this to "IsAvailable" As we are checking if the platform is available for development, not if it is active or not.

XRTK-Core/Packages/com.xrtk.core/Definitions/MixedRealityServiceConfiguration.cs
```diff
@@ -34,14 +35,26 @@ public class MixedRealityServiceConfiguration : IMixedRealityServiceConfiguratio
         /// <param name="instancedType">The concrete type for the <see cref="IMixedRealityService"/>.</param>
         /// <param name="name">The simple, human readable name for the <see cref="IMixedRealityService"/>.</param>
         /// <param name="priority">The priority this <see cref="IMixedRealityService"/> will be initialized in.</param>
-        /// <param name="runtimePlatform">The runtime platform(s) to run this <see cref="IMixedRealityService"/> to run on.</param>
+        /// <param name="runtimePlatforms"></param>
```
StephenHodgson:
```suggestion
        /// <param name="runtimePlatforms">runtimePlatform">The runtime platform(s) to run this <see cref="IMixedRealityService"/> to run on.</param>
```

XRTK-Core/Packages/com.xrtk.core/Services/MixedRealityToolkit.cs
```diff
@@ -779,33 +825,58 @@ private static bool RegisterService(Type interfaceType, IMixedRealityService ser
...
 
+        private static readonly IMixedRealityPlatform[] AllPlatforms = { new AllPlatforms() };
+
         /// <summary>
         /// Creates a new instance of a service and registers it to the Mixed Reality Toolkit service registry for the specified platform.
         /// </summary>
         /// <typeparam name="T">The interface type for the system to be registered.</typeparam>
         /// <param name="concreteType">The concrete type to instantiate.</param>
-        /// <param name="supportedPlatforms">The runtime platform to check against when registering.</param>
+        /// <param name="runtimePlatforms"></param>
```
StephenHodgson:
```suggestion
        /// <param name="runtimePlatforms">The runtime platform to check against when registering.</param>
```

XRTK-Core/Packages/com.xrtk.core/Definitions/MixedRealityServiceConfiguration.cs
```diff
@@ -84,15 +97,48 @@ public uint Priority
             internal set => priority = value;
         }
 
-        [EnumFlags]
         [SerializeField]
-        private SupportedPlatforms runtimePlatform;
+        private RuntimePlatformEntry platformEntries = new RuntimePlatformEntry();
+
+        [NonSerialized]
```
FejZa, line 103: Private fields are non serialized by default, aren't they?

XRTK-Core/Packages/com.xrtk.core/Definitions/Platforms/AndroidPlatform.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) XRTK. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace XRTK.Definitions.Platforms
+{
+    /// <summary>
+    /// Used by the XRTK to signal that the feature is available on the Android platform.
+    /// </summary>
+    public class AndroidPlatform : BasePlatform
```
FejZa, line 9: I see `AndroidPlatform` is defined in Core but `LuminPlatform` in the Lumin module. Wouldn't it make sense to define all the possible XRTK platforms in Core?

XRTK-Core/Packages/com.xrtk.core/Definitions/Platforms/IOSPlatform.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) XRTK. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace XRTK.Definitions.Platforms
+{
+    /// <summary>
+    ///  Used by the XRTK to signal that the feature is available on the iOS platform.
+    /// </summary>
+    public class IOSPlatform : BasePlatform
```
FejZa, line 9: See comment above. It's kind of random that most platforms are defined in Core and some not.

XRTK-Core/Packages/com.xrtk.core/Definitions/Platforms/AndroidPlatform.cs
StephenHodgson: That's actually the benefit of this change so we don't have to define everything in the core

XRTK-Core/Packages/com.xrtk.core/Definitions/Platforms/IOSPlatform.cs
StephenHodgson: We might move these out as we get platform specific sdks.  I thought it was fine for now since we don't have an iOS submodule

XRTK-Core/Packages/com.xrtk.core/Definitions/MixedRealityServiceConfiguration.cs
StephenHodgson: Correct, but the object value doesn't get cleared when the editor recompiles and leaves the last object value from the prev domain reload event. Adding the `NonSerialized` attribute here makes sure that value is cleared on each domain reload.

PR #423, opened 2019-12-22
cleaned up base service

StephenHodgson:

## Overview

Added some docs and removed some trailing spaces.


SimonDarksideJ:
This is including a submodule change, should only be done once the new 144 update is merged and then retargetted

StephenHodgson:
@SimonDarksideJ right this was to test to validate that it was the Oculus submodule that was causing the builds to fail.

SimonDarksideJ:
Needs updating to the latest development branch commit

PR #426, opened 2019-12-27
Change requests/diagnostics system update

StephenHodgson:

## Overview

I refactored the events to match the pattern found in the rest of the toolkit.

## Target of the change:

Is this enhancement for:

## TODO:

- [x] Just need to finish up the actual text assignments on the console window
- [x] Ensure pin/unpin button works


StephenHodgson:
> One thing I don't like about this solution is how in the end, everything is again piped through the MixedRealityDiagnosticsSystem. 

Well something has to do the event routing. 

> The previous implementation allows to simply create a new diagnostics data provider and it's respective handler and then you'd just have to register it the diagnostics system in the inspector and it's ready. Now after creating a new type of data provider you basically have to modify the diagnostics system itself to support the events you need for the new data provider. Makes sense?

Not exactly. We want developers to know what kind of events to expect from these systems. That's mainly why we utilize the interfaces and profiles, to help define the shape of the API and navigate through the implementation to where the problem exists when debugging by following the proper flow of the data from low level data provider, to the platform agnostic service, then to the applications UX.

StephenHodgson:
> I see some cases where fields and properties appear in between methods / constructor etc. I usually prefer having private fields, const etc. then properties, then constructor and then methods. Is that not a convention in XRTK?

Mainly I've been organizing the classes by their interface implementations using regions

FejZa:
Alright. Thanks for clarifying. Are we going to wait for @SimonDarksideJ  to review before merging this into my feature branch?

FejZa:
Nevermind. Forgot you have some open Todos.

StephenHodgson:
Please wait to merge this after I've completed the two items in my todo, thanks :)

StephenHodgson:
I've also got some changes I should probably break out into separate PRs as well

PR #428, opened 2019-12-28
broke out the changes from the diagnostics PR

StephenHodgson:

## Overview

Broke out a few of the braking changes from the Diagnostics PR.

Mainly this is just a cleanup pass.

## Target of the change:

Is this enhancement for:

## Changes:

Some of the biggest changes:

- Updated the `MixedRealityToolkit` service registration to no longer check if a profile is null, as these services require the profiles they're checking. (in fact the `IsServiceEnabled` flag is typically bundled with this profile null check anyway.)
- updated a few of the headers to be XRTK as they're probably nothing like the MRTK counterpart.
- Misc formatting and style changes.
  - naming
  - accessors

## Breaking Changes:

- `IMixedRealityBaseEventSystem.EventListeners` is now a readonly type: `IReadonlyList<GameObject>`

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Services/MixedRealityToolkit.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Services/MixedRealityToolkit.cs
@@ -131,6 +131,7 @@ namespace XRTK.Services
 
         #region Mixed Reality runtime service registry
 
+        // ReSharper disable once InconsistentNaming
         private static readonly Dictionary<Type, IMixedRealityService> activeSystems = new Dictionary<Type, IMixedRealityService>();
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


XRTK-Core/Packages/com.xrtk.core/Services/MixedRealityToolkit.cs
```diff
@@ -131,6 +131,7 @@ public void ResetConfiguration(MixedRealityToolkitConfigurationProfile profile)
 
         #region Mixed Reality runtime service registry
 
+        // ReSharper disable once InconsistentNaming
```
FejZa: I don't like these Resharper comments here. Not a tool everyone uses. But I can live with it if you guys kept them so far.

StephenHodgson: Good thing they're just comments 😉

PR #441, opened 2020-01-12
Pull UpdateController to interface

FejZa:

## Overview

Forgot to pull `UpdateController` to `IMixedRealityController` interface as well in previous PR. `UpdateController` is used to update a controllers state and implemented in `BaseControlller`. Usually overriden to add platform specifics if needed.

## Target of the change:

- Core

## Changes:

- Pulled `UpdateController` to `IMixedRealityController` interface

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Interfaces/Providers/Controllers/IMixedRealityController.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Interfaces/Providers/Controllers/IMixedRealityController.cs
@@ -80,5 +80,10 @@ namespace XRTK.Interfaces.Providers.Controllers
         /// <param name="glbData">The raw binary glb data of the controller model, typically loaded from the driver.</param>
         /// <returns>True, if controller model is being properly rendered.</returns>
         void TryRenderControllerModel(Type controllerType, byte[] glbData = null);
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


StephenHodgson:
What about when some controllers pass through a platform specific parameter through their update methods? (Some platforms do this in their concrete implementations)

FejZa:
That's a good point. I was hoping all controllers would be the same but such is life. In WMR I did the following for a try in that case:

I created 
```
public interface IWindowsMixedRealityCintroller : IMixedRealityController
{
void UpdateController(SpatialIInputSourceState state);
}
```
To extend the default interface for platform specifics. Then inside `UpdateController(...)` I would just store the passed data in a private field and make sure to then call the core interface `UpdateController` snxw it now has all the data to work with. 

What do you think?

FejZa:
I mean the above approach would even work without having UpdateController in the core interface. But it think it does make sense to have a declared way of updating a controller in the interface.

StephenHodgson:
Then we probably need to add this call in each of the classes that don't implement it. It's like to see the PRs for those as well bundled here

StephenHodgson:
It just seems to me to be an implementation detail that doesn't really belong in the controller interface

FejZa:
Agreed. I am not convinced my this one as well anymore.

PR #455, opened 2020-02-17
Introduced IMixedRealityServiceConfiguration

StephenHodgson:
# XRTK - Mixed Reality Toolkit Change Request

## Overview

This interface is meant to be used with serialized structs that define valid `IMixedRealityService` configurations.

This is a step to simplifying the service locator's initial service registration in [`MixedRealityToolkit.InitializeServiceLocator`](https://github.com/XRTK/XRTK-Core/blob/development/XRTK-Core/Packages/com.xrtk.core/Services/MixedRealityToolkit.cs#L333)

This also condenses all of the service lists to use the same base inspector.

## Target of the change:

Is this enhancement for:

- Core (core framework, interfaces and definitions)

## Changes:

- Updated all the existing configurations to implement the new `IMixedRealityServiceConfiguration`

## Notes:
- I intentionally didn't rename the fields bc I didn't want to break backwards compatibility. 
    - one of the great examples of why encapsulation is important!

## TODO:
- [x] Properly filter out the services by interface in the system type drawer dropdowns
- [x] Assertions on registration verifying all the service profiles now implement this interface.

PR #461, opened 2020-03-09
New array Extension plus tests

SimonDarksideJ:

## Overview

* Added new "InitialiseArray" array extension to set all values of an array to a new value, including Tests
* Added null check on Array AddItem
* Fixed license in Core test 2

### DIFF
--- a/XRTK-Core/Assets/XRTK.Tests/Core/TestFixture_03_ExtensionsTests.cs
+++ b/XRTK-Core/Assets/XRTK.Tests/Core/TestFixture_03_ExtensionsTests.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) XRTK. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using NUnit.Framework;
+using UnityEngine;
+using XRTK.Extensions;
+
+namespace XRTK.Tests.Core
+{
+    public class TestFixture_03_ExtensionsTests
+    {
+        [Test]
+        public void Test_01_Array_AddItem_Extension()
+        {
+            Vector2[] testArray = null;
+            Vector2[] newArray =  null;
+            try
+            {
+                newArray = testArray.AddItem(Vector2.zero);
+            }
+            catch { }
+
+            Assert.IsNotNull(newArray);
+
+            testArray = new Vector2[5];
+            newArray = testArray.AddItem(Vector2.one);
+
+            Assert.IsTrue(newArray[newArray.Length - 1] == Vector2.one);
+            Assert.IsTrue(newArray.Length > testArray.Length);
+            Assert.IsTrue(newArray.Length - testArray.Length == 1);
+        }
+
+        [Test]
+        public void Test_02_Array_InitialiseArray_Extension()
+        {
+            Vector2[] testArray = null;
+            Vector2[] newArray = null;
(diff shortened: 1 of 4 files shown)


XRTK-Core/Assets/XRTK.Tests/Core/TestFixture_03_ExtensionsTests.cs
```diff
@@ -0,0 +1,58 @@
...
+using NUnit.Framework;
+using UnityEngine;
+using XRTK.Extensions;
+
+namespace XRTK.Tests.Core
+{
+    public class TestFixture_03_ExtensionsTests
+    {
+        [Test]
+        public void Test_01_Array_AddItem_Extension()
```
StephenHodgson: Can we also test to add the item at an random index and ensure that it's inserted in the correct postion?

XRTK-Core/Packages/com.xrtk.core/Extensions/ArrayExtensions.cs
```diff
@@ -37,12 +37,43 @@ public static bool IsValidArray(this Array array)
         /// <param name="array">The array to extend</param>
         /// <param name="newItem">The item to add to the array</param>
         /// <returns></returns>
-        public static T[] AddItem<T>(this T[] array, T newItem, int insertAtIndex = 0)
+        public static T[] AddItem<T>(this T[] array, T newItem)
         {
+            // Initialise the array if it is null
+            if (array == null)
+            {
+                array = new T[0];
```
StephenHodgson: Take a look at my changed in [455](https://github.com/XRTK/XRTK-Core/pull/455/files?file-filters%5B%5D=.cs&hide-deleted-files=true#diff-d33dfd0c48086e9a05bb9888d60a437bR45)
```suggestion
                return new[] { newItem };
```

XRTK-Core/Packages/com.xrtk.core/Extensions/ArrayExtensions.cs
```diff
@@ -37,12 +37,43 @@ public static bool IsValidArray(this Array array)
         /// <param name="array">The array to extend</param>
         /// <param name="newItem">The item to add to the array</param>
         /// <returns></returns>
-        public static T[] AddItem<T>(this T[] array, T newItem, int insertAtIndex = 0)
```
StephenHodgson: I don't mind keeping this if we're actually inserting it correctly. ATM it doesn't actually insert it at the right index, which is a bug, and why I modified it in the other PR I was working on.

XRTK-Core/Assets/XRTK.Tests/Core/TestFixture_03_ExtensionsTests.cs
SimonDarksideJ: This is not an option it currently supports. Might add that in the future

XRTK-Core/Packages/com.xrtk.core/Extensions/ArrayExtensions.cs
SimonDarksideJ: We removed the insert at index option for now

SimonDarksideJ:
Applied suggested fix to both extensions, if the array is null, return a single item array with the item provided

PR #473, opened 2020-03-28
Fixed some issues with registering extension services and data providers

StephenHodgson:
# XRTK - Mixed Reality Toolkit Change Request

## Overview

Did a bunch more work to fix some issues around registering extension services and data providers.

PR #474, opened 2020-03-29
Enabled specific profile type filtering in the configuration list

StephenHodgson:

## Overview

Enabled the ability to specify specific profile types in the profile object picker based on the service's constructor profile type.

image
image


StephenHodgson:
Blocked by #473

StephenHodgson:
We are moving towards a clone first approach, but some profiles aren't required, so having the ability to create new profile instances via editor is still an important feature to have.

PR #480, opened 2020-03-30
updated simulated hand data profile with required fields for IMixedRealityHandControllerDataProvider

StephenHodgson:

## Overview

This should get the branch to compile successfully again.
All I did was add the required fields to the profile.


StephenHodgson:
Yeah same concern here as well about duplicating them, but atm I don't see a way around it.

Goes back to the discussion we were having around global vs platform specific settings.

PR #216, opened 2019-06-21
Feature - Hand Controllers + Simulation - CORE

FejZa:
# XRTK - Mixed Reality Toolkit Change Request

## Overview

This PR adds the Core components for hand controllers in XRTK. Additionally it adds the concept of controller simulation with a first implementation of simulated hand controllers. It provides a base for further hand related implementations, such as pointers and UX.

### Screenshots

game_view
new_sim_profile
dp_added
mapping

## Target of the change:

- Core (core framework, interfaces and definitions)
- SDK (default profiles and assets)

## Submodule PRs

- https://github.com/XRTK/SDK/pull/45
- https://github.com/XRTK/Lumin/pull/49
- https://github.com/XRTK/Oculus/pull/44
- https://github.com/XRTK/WindowsMixedReality/pull/58

PR #495, opened 2020-04-04
ported game object extension from #216

StephenHodgson:

## Overview

- Adds `gameObject.GetOrAddComponet<Component>()`


FejZa:
Worth noting that as soon as we are on 2019.2+ Unity has `TryGetComponent` and others built-in. But not GetOrAdd. That one will still remain useful eventually.

PR #496, opened 2020-04-04
cleaned up base hand controller

StephenHodgson:


StephenHodgson:
> Really struggling with all those var changes. But I understand I am kinda a unicorn not liking var.

I used to feel the same way lol. I'm always back and forth. I just think less to read is better. 

As long as the field name is descriptive I don't usually care about the `Type`

PR #502, opened 2020-04-06
Fix/android platform usings

FejZa:

Forgot to put the using into a preprocessor directive and some auto cleanup propably removed them.


StephenHodgson:
Just for the sake of keeping it clean, why not just inline the namespace?

FejZa:
Can do no problem

FejZa:
Done

PR #508, opened 2020-04-07
Controller Mapping Profiles Refactor

StephenHodgson:
# XRTK - Mixed Reality Toolkit Change Request

## Overview

Changes outlined in #494

## Target of the change:

Is this enhancement for:

- Core (core framework, interfaces and definitions)

## Submodule Changes

- https://github.com/XRTK/SDK/pull/159
- https://github.com/XRTK/Lumin/pull/67
- https://github.com/XRTK/Oculus/pull/62
- https://github.com/XRTK/WindowsMixedReality/pull/71

## TODO

- [x] Update Interaction Mappings to use Processors
- [x] Update Unit Tests
- [x] Update how we instantiate controllers and pointers in the data provider
- [x] Update Controller popout window
- [x] Regenerate all the default profiles
- [x] Cleanup PR and break out unrelated changes
- [x] Update UX @FejZa 
- [x] Update Controller image paths
- [x] https://github.com/XRTK/XRTK-Core/pull/517

PR #507, opened 2020-04-07
Release 0.2.0

StephenHodgson:
# 🚧 TODO 🚧 

- [x] #524 Service Locator Refactor
- [x] #517 Camera System Refactor
- [x] #504 Spatial Awareness Refactor
- [x] #508 Controller Mapping Refactor
- [x] #549 Cleanup packages to make sure everything lives in the correct place
- [x] #550 CI/CD Pipeline Upgrade
- [x] #554 Add platform profile installer to ensure new platforms get their data providers registered
- [x] #558 Improved UPM package installation
- [x] #559 Fix canvas utility issue
- [x] #561 Fix controller popout NRE
- [x] #562 update package installer to regenerate asset guid references
- [x] #563 update sdk package to use hidden `Prefabs~` folder
- [x] #570 Service Wizard
- [x] #560 Controller near interactions
- [x] #590 Fix camera rig setup on configure
- [x] #591 Fix blank runtime platforms on installed platform configurations
- [x] #552 Hand input and gesture actions
- [X] https://github.com/XRTK/SteamVR/issues/11 Fix OpenVR controller offset
- [x] https://github.com/XRTK/SDK/issues/203 Build errors when validating project assets
- [x] #675 Fix platform configurations
- [x] https://github.com/XRTK/SDK/issues/201 Missing asset files after SDK import via UPM
- [x] #649 Fix NRE when creating new `IMixedRealityCameraRig`
- [x] Update project to latest 2019.4 LTS release
- [x] #700 Symlink step "Updating Submodules" shows up for too long
- [x] #702 Prompt user to install SDK if no configurations found
- [x] #504 Fix teleporting
- [x] https://github.com/XRTK/XRTK-Core/pull/694 Update Spatial Awareness System
- [x] https://github.com/XRTK/WindowsMixedReality/issues/87 Fix NRE when no camera data provider is setup for the platform
- [ ] ~#603 Go fully native~

## :shipit: Release Tasks
- [x] #580 Create a preview upm package every time development branch gets updated when there is an active pull request to the master branch
- [x] #584 Update getting started documentation
- [x] #670 Documents Checklist
- [x] Follow [Test Plans](https://github.com/XRTK/XRTK-Core/blob/development/articles/testplans/TP_Overview.md)
- [x] Test Core installation only
- [x] Test Core & SDK Installation only
- [x] Test Core & all packages

PR #517, opened 2020-04-11
Camera System Refactor

StephenHodgson:

## Overview

Refactored the Camera System to have per platform data providers and settings on top of the global ones.

## Target of the change:

Is this enhancement for:

## Breaking Changes:

- Completely changed the `IMixedRealityCameraSystem`
- Added `IMixedRealityDataProvider`, all of the original logic in the camera system was moved into the `BaseCameraDataProvider` class.
- Updated all the namespaces for the camera system files. (Some classes didn't have a namespace to begin with)

## Submodule Changes

https://github.com/XRTK/SDK/pull/153
https://github.com/XRTK/Oculus/pull/54
https://github.com/XRTK/Lumin/pull/60
https://github.com/XRTK/WindowsMixedReality/pull/66

## TODO

Tests should also include moving the playspace to ensure that both spatial awareness and teleportation is keeping the tracked objects in sync with the playspace.

- [x] #524
- [x] #529
- [x] Test on OpenVR
- [x] Test in editor without headset
- [x] Test on Lumin
    - [x] Test on device
    - [x] Test using Remoting
- [x] Test on WMR
    - [x] Test in editor with VR headset
    - [x] Test on HoloLens in editor using remoting
    - [x] Test on HoloLens device without using remoting
- [x] Test on Oculus
  - [x] Quest on Device
  - [x] Quest in Editor with link cable


XRTK-Core/ProjectSettings/QualitySettings.asset
```diff
@@ -4,7 +4,7 @@
 QualitySettings:
   m_ObjectHideFlags: 0
   serializedVersion: 5
-  m_CurrentQuality: 5
+  m_CurrentQuality: 0
```
StephenHodgson, line 7: revert

SimonDarksideJ:
This PR needs a lot of cleaning up as it contains so many different changes unrelated to the Camera System update

XRTK-Core/Packages/com.xrtk.core/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
```diff
@@ -10,28 +11,31 @@ namespace XRTK.Interfaces.CameraSystem
...
         /// </summary>
-        bool IsOpaque { get; }
+        IReadOnlyCollection<IMixedRealityCameraDataProvider> CameraDataProviders { get; }
 
         /// <summary>
-        /// Is the current camera displaying on a traditional 2d screen or a stereoscopic display?
+        /// The reference to the <see cref="IMixedRealityCameraRig"/> attached to the Main Camera (typically this is the player's camera).
         /// </summary>
-        bool IsStereoscopic { get; }
+        IMixedRealityCameraRig MainCameraRig { get; }
```
SimonDarksideJ, line 21: I don't agree with simply renaming this as it causes so much downstream change for no benefit.
Why change the name?  This is a major breaking change

FejZa:
I like where this PR is going but I am a bit confused about how things are done. It seems the Camera System is now polling it's data providers for data, whereas in other systems data providers are pushing the data to the system.

Why does the camera system have it's own kind of "registry"? Couldn't we get all the `ICameraDataProvider`s simply using `MixedRealityToolkit.GetActiveServices<ICameraDataProvider>`?

StephenHodgson:
> This PR needs a lot of cleaning up as it contains so many different changes unrelated to the Camera System update

Yup that's why I opened #518

XRTK-Core/Packages/com.xrtk.core/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
StephenHodgson: the entire `IMixedRealityCameraSystem` interface changed. I also wanted to make a distinction here. It's possible we may have multiple cameras that have different rig requirements in the future.

StephenHodgson:
> It seems the Camera System is now polling it's data providers for data

I'm confused, I don't see where we're polling for data.

StephenHodgson:
> Why does the camera system have it's own kind of "registry"? Couldn't we get all the `ICameraDataProvider`s simply using `MixedRealityToolkit.GetActiveServices<ICameraDataProvider>`?

It doesn't have it's own registry. In fact it's setup exactly the same way as the Input System.

StephenHodgson:
Still need to implement head height for WMR/OpenVR

StephenHodgson:
Tested this on all the platforms except for HoloLens and Oculus

SimonDarksideJ:
@FejZa will you have time to test HoloLens with this PR.  I know you have a Quest but I can test that if you're busy atm

FejZa:
I can test this on HoloLens 2 yes.

FejZa:
## Tested on HoloLens 2:

### Build and Deploy to device

From what I understand this PR is supposed to do it works. My scene floor at (0,0,0) appeared roughly at my feet when running on device.

### Editor Remoting to Hololens 2

In this case it didn't work. When playing my camera sits on the floor.

### General feedback

- I am not a bit confused by this PR and what it's actually solving.
- What's "EditorBuildTarget"? It's not clear what it's supposed to do and it's not a valid platform. It's very confusing to me.
- Instanced type "BaseCameraDataProvider" is weird. I'd rather have e.g. a `DefaultCameraDataProvider` class.
- I though CameraCache is something we deprecated but it's now being used here?
- I asked this before, why does the camera system profile not make sure of the new services profile
but instead adds a dedicated field for the new profiles?
- I don't get why in this instance having `RegisterCameraDataProvider` and `UnRegisterCameraDataProvider` in the camera system is okay but in other instaces it was against the toolkit's principles. Instead of having data providers register themself to the camera system, why not just do `MixedRealityToolkit.GetActiveServices<IMixedRealityCameraDataProvider>()` in the camera system's `Initalize()` hook.

StephenHodgson:
> - I am not a bit confused by this PR and what it's actually solving.

This pull request to align the camera system to better fit the service system / data provider patterns we're using in other, more stable parts of the architecture. Mainly platform agnostic service with a platform specific provider. 

The camera system is meant to setup, control, and sync the Camera Rigs movements in the virtual space in relation to the camera's physical location.  The Camera data provider should have proper 

> - What's "EditorBuildTarget"? It's not clear what it's supposed to do and it's not a valid platform. It's very confusing to me.

This was unrelated to this PR (See #521). All this does is makes it so that the service only runs in the editor when the editor build target matches. It's opposite is `Editor Only` which enables a service to always run in the editor, but not in the runtime.

> - Instanced type "BaseCameraDataProvider" is weird. I'd rather have e.g. a `DefaultCameraDataProvider` class.

It matches the naming conventions we've used so far. I did update the service's name in the configuration to `DefaultCameraDataProvider`. Does this work for you?

> - I thought `CameraCache` is something we deprecated but it's now being used here?

It was never deprecated. We wanted to use the camera system in favor of the `CamerCache` because there may be rigs or player cameras that aren't tagged as `MainCamera`.

> - I asked this before, why does the camera system profile not make use of the new services profile but instead adds a dedicated field for the new profiles?

Mainly I did this so the inspector wouldn't be cluttered since we also have the global overrides in the main system profile. This is how the Input System profile is currently also setup.

> - I don't get why in this instance having `RegisterCameraDataProvider` and `UnRegisterCameraDataProvider` in the camera system is okay but in other instaces it was against the toolkit's principles. Instead of having data providers register themself to the camera system, why not just do `MixedRealityToolkit.GetActiveServices<IMixedRealityCameraDataProvider>()` in the camera systems `Initalize()` hook.

Because we've added the `serviceParent` to the constructor for all data providers we no longer have to use `GetActiveServices` to understand which service this data provider belongs to. But yeah this is something I was looking at revisiting.

The main reason why I was doing this was bc I kept running into a race condition with the spatial awareness system and boundary systems that depends on the rig being properly setup by the time they are initialized and enabled. A lot of these problems are addressed in #524

StephenHodgson:
> but in other instances it was against the toolkit's principles.

I'm not sure what you mean by this. All the controller data providers register themselves with the input system once they're found, using events. But in this case I felt like using events was a lot of overhead for something simple.

StephenHodgson:
> ### Editor Remoting to Hololens 2
>
>In this case it didn't work. When playing my camera sits on the floor.

@FejZa Is the playspace moved down by the default head height, or is it at the origin?

FejZa:
It's at the origin

StephenHodgson:
> It's at the origin

Can you give me the relative positions of the playspace, camera, and the body transform positions?

FejZa:
MixedRealityPlayspace is at
Pos: 0,0,0
Rot: 0,0,0

MainCamera is at:
Pos: Roughly 0,0,0 since the HoloLens is tracking movement
Rot: Roughly 0,0,0 as well

PlayerBody is at:
Pos: x and z roughly 0 and y -1.57254 which is roughly the configured default head heigth I guess

FejZa:
Does that help?

StephenHodgson:
We probably need to override the reset rig method in the wmr implementation

StephenHodgson:
@FejZa let me know if this change fixed the issue you had in the editor remoteing

FejZa:
Hi @StephenHodgson,

no it didn't. The change you made is overriding `ResetRigTransforms` in the WMR data provider, right?
I created a recording for you to see what's going on:
https://1drv.ms/v/s!ApWA4y9RKCvzrt0sGPoMRZoCxuOeYg?e=n2XWpL

I was able to achieve what I think is the right behaviour by changing:
`CameraRig.PlayspaceTransform.position = Vector3.zero;`

to

`CameraRig.PlayspaceTransform.position = IsStereoscopic || !IsOpaque ? new Vector3(0f, HeadHeight, 0f) : Vector3.zero;`

The issue is, that no matter what you do to the camera transform, on HoloLens it will get overriden by the system's head tracking. So in order to offset the camera, you have to offset it's parent. But I don't think moving the playspace up is what you'd like to see.

SimonDarksideJ:
We should most definitely NOT be moving the playspace.
Either we move the camera to head height, or the device/api should, but we shouldn't be making other alterations really.  Unless I'm misunderstanding the problem here and why it's now an issue.

FejZa:
The issue is that while on other platforms the device actually moves the camera to head height, the HoloLens makes the head position at launch be the the origin of the scene (0,0,0). That way anything you have in your scene at height 0 will be at head height instead of "on the floor". 

But I agree moving the playspace doesn't seem like a solution to this. Maybe on HoloLens we just can't achieve the same effect as with other platforms and we just gotta stick with it.

Never really understood the point of the name `MixedRealityPlayspace` though. I know many people are confused by it's naming because it suggests you need to put your custom scene content into it. Which is not the case. But that's another discussion.

StephenHodgson:
> We should most definitely NOT be moving the playspace.

Why not? I'm pretty sure that's the only thing we should be moving. We def shouldn't be moving the camera. The playspace is what moves we we need to synchronize the virtual to the physical space. 

> Unless I'm misunderstanding the problem here and why it's now an issue.

I think this is the case. We're only trying to fix the problem with the changes with the in editor remoting. Everything else works as it did previously.

> The issue is that while on other platforms the device actually moves the camera to head height, the HoloLens makes the head position at launch be the the origin of the scene (0,0,0). That way anything you have in your scene at height 0 will be at head height instead of "on the floor".

It's not the device doing this, but the Unity Editor. There's an XR subsystem under the hood that's moving the camera's for each device to some predetermined location. When this system is complete we'll no longer be using that but adjusting using each data provider's api. (that change will be made as we move closer from breaking away from `PoseDeriver`, I was already having a similar issue with Lumin, but decided that I could revisit it later)

> Never really understood the point of the name `MixedRealityPlayspace` though. I know many people are confused by its naming because it suggests you need to put your custom scene content into it. Which is not the case. But that's another discussion.

The playspace is a virtual representation of they players physical space. It helps synchronize your frame of reference with the virtual environment you're in. All transforms attached to this object are considered to be synchronized to the player's local frame of reference. This includes things like the controllers, the spatial mesh, and any other physically tracked object that needs to be represented in the virtual space. 

Things outside of this playspace should contain the purely virtual elements of the scene and possibly even shared content.

We HAVE to move the playspace to do multiple things, like teleportation, moving the player camera in general, or starting users at a fixed location when the application starts. (Even Unity docs say that you have to parent a camera under a GameObject, and not to move the camera directly).

The only reason we make sure everything is synchronized at the origin is specifically for Windows Mixed Reality VR headsets bc then all of the tracked objects in the users local frame of reference is offset by that same amount (but I highly suspect that's bc of a Unity bug). We've documented this in the old days for MRTK and is something we've continued to do for that reason. 

In either case, Thanks Dino for giving me a better idea of what's going on. If I had a device I wouldn't have to do more pecking and hunting to get this right (although a PR would be nice).

Hopefully in the future we can break all the dependencies of the PoseDriver and Unit AutoXRCameraAdjustment apis, but this PR is one of the first steps in that direction.

StephenHodgson:
@FejZa I'm wondering why the in editor remoting works differently than on device. 

The PR works correctly on device, but not using remoting?

StephenHodgson:
Okay I went back and reworked some things. Let's see how this build works

SimonDarksideJ:
Can we please ensure that PR's across submodules are all in the same folder and have the same name. This one is far too confusing to checkout fully

StephenHodgson:
That was my attempt to start putting things into their correct locations, but I guess I can revert that. Just more work to have to go back and do it again later.

StephenHodgson:
What exactly is confusing?

SimonDarksideJ:
🤷‍♂️
* fix/camera-system
* feature/camera-system-refactor
* dev/camera-system-refactor
* dev/camera-system-refactor
* dev/camera-system

What could possibly be confusing 😜

StephenHodgson:
Haha, ah yeah the branch names. Yeah I could have done better there. I've just been using the links to navigate back and forth.

StephenHodgson:
@SimonDarksideJ Just a heads up. I was running into issues with Oculus eye height, and temporarily removed it. I wanted to add a `OculusCameraDataProvider` as well, but we also need to add the platform definition for it. The api's were overriding the height for other platforms like Lumin during the editor simulation. I still wanna figure out what kind of platform definition we need for Oculus Specifically.

SimonDarksideJ:
#facepalm intentionally breaking the one thing we added for Oculus because Unity didn't do its duty 🤦‍♂️

FejZa:
What an engine's duty is very relative and "depends". I am with Stephen here, setting the eye height always seemed to me like something that is not in the responsibility of the controller data provider and had a "workaround / hack" feel to it.

SimonDarksideJ:
Oh Agreed @FejZa , but at this time we don't have platform-specific PlatformCameraProviders, we still rely on Unity for that.

FejZa:
I think we'll get there 😊. @StephenHodgson did you do any more changes here? Because it seems it's working now.

FejZa:
@StephenHodgson I did some more testing and everything is working now, Editor Remoting and deployed to device.

SimonDarksideJ:
Tested on a few platforms and is working well, quick UX request, can we add a fold out for the camera providers to simplify the UI
image

Just place it below the profile property (hide foldout if no profile assigned (or is that even allowed?)

SimonDarksideJ:
Additionally, The project will run without a camera profile, but the console is flooded with errors and no controllers appear (so are controllers now tightly bound to a specific camera, or any camera)

This should fail better.

StephenHodgson:
@FejZa yeah more work to get Oculus working again

StephenHodgson:
@SimonDarksideJ I'll take another pass at the issues you brought up

StephenHodgson:
@FejZa just to confirm this works for HoloLens on device and remoting from the editor?

FejZa:
I hereby confirm it works on device and remoting from editor.

StephenHodgson:
@SimonDarksideJ those errors you're seeing are happening in development as well. Seems as though the Camera System is always required atm.

StephenHodgson:
It's bc the GazeProvider expects that the camera be parented under another GameObject but it's camera parent is null when we disable the camera system

StephenHodgson:
Fixed the NREs in https://github.com/XRTK/SDK/pull/156

StephenHodgson:
>  so are controllers now tightly bound to a specific camera, or any camera

Pretty sure their visualizations are. Without a camera how'd you even see them? lol

Besides afaik the controllers were already working without the input system? It's the UX pointer visualizations and the like that require both the Input System and Camera System. Let's revisit this after #508

StephenHodgson:
@SimonDarksideJ cleaned up the camera system profile UX
image

StephenHodgson:
@SimonDarksideJ @FejZa this is ready for review for Oculus implementation, but there are PRs that block this one.

SimonDarksideJ:
@FejZa where did you get up to with your testing on this. Got to admit, I'm getting a little lost with all the breakout PR's now

FejZa:
Its tested and working on HoloLens. But the PR is blocked since I can't test on Oculus Quest currently because Quest is totally broken currently die to another bug. Once we fix that and I can verify this works as intended ok quest we're good.

StephenHodgson:
@SimonDarksideJ all the breakout PRs target the branches they're broken out from. 

This is the base PR at the moment afaik.

- Camera Refactor
  - Spatial Awareness Refactor
  - Controller Mapping Refactor

StephenHodgson:
Check https://github.com/XRTK/XRTK-Core/pull/507 for the order in which these PRs need to be merged.

XRTK-Core/Packages/com.xrtk.core/Services/TeleportSystem/MixedRealityTeleportSystem.cs
```diff
@@ -213,12 +213,17 @@ private void ProcessTeleportationRequest(TeleportEventData eventData)
...
+                : MixedRealityToolkit.CameraSystem.MainCameraRig.CameraTransform;
             var cameraPosition = cameraTransform.position;
             var cameraParent = cameraTransform.parent;
-            Debug.Assert(cameraParent != null, "Teleport system requires that the camera be parented under another object.");
+            Debug.Assert(cameraParent != null, "The Teleport System requires that the camera be parented under another object.");
             var height = targetPosition.y;
 
+            if (MixedRealityToolkit.CameraSystem != null)
+            {
+                height += MixedRealityToolkit.CameraSystem.MainCameraRig.PlayspaceTransform.position.y;
```
StephenHodgson, line 224: We can probably just use the cameraTransform.parent.position.y so we don't have to check for the camera system again.

XRTK-Core/Packages/com.xrtk.core/Interfaces/CameraSystem/IMixedRealityCameraDataProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>
+    /// Base interface for implementing camera data providers to be registered with the <see cref="IMixedRealityCameraSystem"/>
+    /// </summary>
+    public interface IMixedRealityCameraDataProvider : IMixedRealityDataProvider
+    {
+        /// <summary>
+        /// Is the current camera displaying on an Opaque (AR) device or a VR / immersive device
+        /// </summary>
+        bool IsOpaque { get; }
```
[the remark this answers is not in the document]
StephenHodgson: It is possible we might have different cameras that have opaque vs transparent rendering qualities. I think it's best left on the device level.

XRTK-Core/Packages/com.xrtk.core/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
```diff
@@ -10,28 +11,31 @@ namespace XRTK.Interfaces.CameraSystem
...
+        /// <param name="value"></param>
+        void SetHeadHeight(float value);
 
         /// <summary>
-        /// The default head height when a platform doesn't automatically set it.
+        /// Registers the <see cref="IMixedRealityCameraDataProvider"/> with the <see cref="IMixedRealityCameraSystem"/>.
         /// </summary>
-        float DefaultHeadHeight { get; }
+        /// <param name="dataProvider"></param>
+        void RegisterCameraDataProvider(IMixedRealityCameraDataProvider dataProvider);
```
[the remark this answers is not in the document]
StephenHodgson: I think it was mainly bc of the event handlers we are using. 

We're not using event handling for anything in the camera system afaik.

XRTK-Core/Packages/com.xrtk.core/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -312,32 +289,49 @@ private void ApplySettingsForOpaqueDisplay()
...
-            bodyLocalPosition.z = cameraPosition.z;
+            bodyLocalPosition.x = cameraLocalPosition.x;
+
+            if (HeadHeight > 0f)
+            {
+                bodyLocalPosition.y = cameraLocalPosition.y - HeadHeight;
+            }
+            else
+            {
+                bodyLocalPosition.y = cameraLocalPosition.y - 1.6f;
```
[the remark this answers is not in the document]
StephenHodgson: Yeah we probably shouldn't use a magic number here. It is supposed to be the default head height but that may be zero in some cases and it won't properly offset the body position. I think that's why I hard coded it here.

XRTK-Core/Packages/com.xrtk.core/Definitions/CameraSystem/MixedRealityCameraSystemProfile.cs
```diff
@@ -0,0 +1,156 @@
...
+using UnityEngine;
+using XRTK.Attributes;
+using XRTK.Definitions.Utilities;
+using XRTK.Services.CameraSystem;
+using XRTK.Interfaces.CameraSystem;
+
+namespace XRTK.Definitions.CameraSystem
+{
+    /// <summary>
+    /// This <see cref="BaseMixedRealityProfile"/> tells you if your head mounted display (HMD)
```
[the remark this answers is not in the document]
StephenHodgson: Think this was just a renamed file. We can update the language

XRTK-Core/Packages/com.xrtk.core/Definitions/CameraSystem/MixedRealityCameraSystemProfile.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// is a transparent device or an occluded device.
+    /// Based on those values, you can customize your camera and quality settings.
+    /// </summary>
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Camera System Profile", fileName = "MixedRealityCameraSystemProfile", order = (int)CreateProfileMenuItemIndices.Camera)]
+    public class MixedRealityCameraSystemProfile : BaseMixedRealityProfile
```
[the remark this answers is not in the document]
StephenHodgson: [it is in the latest version](https://github.com/XRTK/XRTK-Core/pull/517/files#diff-2442037e2805bcead939230d2ad0a468R16). You're looking at an older commit?

XRTK-Core/Packages/com.xrtk.core/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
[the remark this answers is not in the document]
StephenHodgson: Idk I can go either way with this. 

I'll look back at the Input and Diagnostics systems to see if we can do something similar.

The biggest difference is that the Diagnostics and Input Systems are Event Systems, while the Camera System is not. No one is listening for events here.

XRTK-Core/Packages/com.xrtk.core/Inspectors/Profiles/BaseMixedRealityCameraDataProviderProfileInspector.cs
```diff
@@ -51,16 +51,17 @@ protected override void OnEnable()
             bodyAdjustmentSpeed = serializedObject.FindProperty(nameof(bodyAdjustmentSpeed));
         }
 
+
```
StephenHodgson, line 54:
```suggestion
```

XRTK-Core/Packages/com.xrtk.core/Definitions/Controllers/Simulation/Hands/SimulatedHandControllerPose.cs
```diff
@@ -151,7 +151,7 @@ private static void OffsetJointsRelativeToOpenPosePalmPosition(SimulatedHandCont
         /// </summary>
         public void ComputeJointPoses(Handedness handedness, Quaternion rotation, Vector3 position, MixedRealityPose[] jointsOut)
         {
-            Quaternion cameraRotation = MixedRealityToolkit.CameraSystem.CameraRig.PlayerCamera.transform.rotation;
+            Quaternion cameraRotation = MixedRealityToolkit.CameraSystem.MainCameraRig.PlayerCamera.transform.rotation;
```
StephenHodgson, line 154: We should check if the camera system is null here and fall back to `CameraCache.Main`

XRTK-Core/Packages/com.xrtk.core/Definitions/Controllers/Simulation/Hands/SimulatedHandControllerPose.cs
```diff
@@ -185,7 +185,7 @@ public void ComputeJointPoses(Handedness handedness, Quaternion rotation, Vector
         public void ParseFromJointPoses(MixedRealityPose[] joints, Handedness handedness, Quaternion rotation, Vector3 position)
         {
             var invRotation = Quaternion.Inverse(rotation);
-            var invCameraRotation = Quaternion.Inverse(MixedRealityToolkit.CameraSystem.CameraRig.PlayerCamera.transform.rotation);
+            var invCameraRotation = Quaternion.Inverse(MixedRealityToolkit.CameraSystem.MainCameraRig.PlayerCamera.transform.rotation);
```
StephenHodgson, line 188: We should check if the camera system is null here and fall back to `CameraCache.Main`

XRTK-Core/Packages/com.xrtk.core/Inspectors/Profiles/BaseMixedRealityCameraDataProviderProfileInspector.cs
```diff
@@ -29,7 +30,6 @@ public class MixedRealityCameraProfileInspector : BaseMixedRealityProfileInspect
 
         private readonly GUIContent nearClipTitle = new GUIContent("Near Clip");
         private readonly GUIContent clearFlagsTitle = new GUIContent("Clear Flags");
-
```
StephenHodgson, line 32: put this line back

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseController.cs
```diff
@@ -246,7 +246,7 @@ internal async Task TryRenderControllerModelAsync(Type controllerType, byte[] gl
                 if (useSystemDefaultModels && gltfObject != null)
                 {
                     controllerModel.name = $"{controllerType.Name}_Visualization";
-                    controllerModel.transform.SetParent(MixedRealityToolkit.CameraSystem?.CameraRig.PlayspaceTransform);
+                    controllerModel.transform.SetParent(MixedRealityToolkit.CameraSystem?.MainCameraRig.PlayspaceTransform);
```
StephenHodgson, line 249: We should check if the camera system is null here and fall back to `CameraCache.Main`

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseController.cs
```diff
@@ -255,7 +255,7 @@ internal async Task TryRenderControllerModelAsync(Type controllerType, byte[] gl
                 //If the model was a prefab
                 else
                 {
-                    var controllerObject = UnityEngine.Object.Instantiate(controllerModel, MixedRealityToolkit.CameraSystem?.CameraRig.PlayspaceTransform);
+                    var controllerObject = UnityEngine.Object.Instantiate(controllerModel, MixedRealityToolkit.CameraSystem?.MainCameraRig.PlayspaceTransform);
```
StephenHodgson, line 258: We should check if the camera system is null here and fall back to `CameraCache.Main`

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseController.cs
StephenHodgson, line 249: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseController.cs
StephenHodgson, line 258: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/BaseControllerDataProvider.cs
```diff
@@ -45,7 +45,7 @@ protected virtual IMixedRealityPointer[] RequestPointers(SystemType controllerTy
                     if (((useSpecificType && pointerProfile.ControllerType.Type == controllerType.Type) || (!useSpecificType && pointerProfile.ControllerType.Type == null)) &&
                         (pointerProfile.Handedness == Handedness.Any || pointerProfile.Handedness == Handedness.Both || pointerProfile.Handedness == controllingHand))
                     {
-                        var pointerObject = Object.Instantiate(pointerProfile.PointerPrefab, MixedRealityToolkit.CameraSystem?.CameraRig.PlayspaceTransform);
+                        var pointerObject = Object.Instantiate(pointerProfile.PointerPrefab, MixedRealityToolkit.CameraSystem?.MainCameraRig.PlayspaceTransform);
```
StephenHodgson, line 48: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/Simulation/Hands/SimulatedHandController.cs
```diff
@@ -328,9 +328,9 @@ private void HandleSimulationInput(MixedRealityPose rootPoseDelta)
             // Apply mouse delta x/y in screen space, but depth offset in world space
             screenPosition.x += rootPoseDelta.Position.x;
             screenPosition.y += rootPoseDelta.Position.y;
-            Vector3 newWorldPoint = MixedRealityToolkit.CameraSystem.CameraRig.PlayerCamera.ScreenToWorldPoint(ScreenPosition);
-            newWorldPoint += MixedRealityToolkit.CameraSystem.CameraRig.PlayerCamera.transform.forward * rootPoseDelta.Position.z;
-            screenPosition = MixedRealityToolkit.CameraSystem.CameraRig.PlayerCamera.WorldToScreenPoint(newWorldPoint);
+            Vector3 newWorldPoint = MixedRealityToolkit.CameraSystem.MainCameraRig.PlayerCamera.ScreenToWorldPoint(ScreenPosition);
+            newWorldPoint += MixedRealityToolkit.CameraSystem.MainCameraRig.PlayerCamera.transform.forward * rootPoseDelta.Position.z;
+            screenPosition = MixedRealityToolkit.CameraSystem.MainCameraRig.PlayerCamera.WorldToScreenPoint(newWorldPoint);
```
StephenHodgson, line 333: We should check if the camera system is null here and fall back to `CameraCache.Main`

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/Simulation/Hands/SimulatedHandController.cs
```diff
@@ -349,7 +349,7 @@ private void UpdatePoseFrame()
 
             currentPoseBlending = TargetPoseBlending;
             var rotation = Quaternion.Euler(HandRotateEulerAngles);
-            var position = MixedRealityToolkit.CameraSystem.CameraRig.PlayerCamera.ScreenToWorldPoint(ScreenPosition + JitterOffset);
+            var position = MixedRealityToolkit.CameraSystem.MainCameraRig.PlayerCamera.ScreenToWorldPoint(ScreenPosition + JitterOffset);
```
StephenHodgson, line 352: We should check if the camera system is null here and fall back to `CameraCache.Main`

XRTK-Core/Packages/com.xrtk.core/Services/BoundarySystem/MixedRealityBoundarySystem.cs
```diff
@@ -323,7 +323,7 @@ private GameObject BoundaryVisualizationParent
                 }
 
                 var visualizationParent = new GameObject("Boundary System Visualizations");
-                visualizationParent.transform.parent = MixedRealityToolkit.CameraSystem?.CameraRig.PlayspaceTransform;
+                visualizationParent.transform.SetParent(MixedRealityToolkit.CameraSystem?.MainCameraRig.PlayspaceTransform, false);
```
StephenHodgson, line 326: This is fine as it'll set the parent to nothing, which is fine.

StephenHodgson, line 326: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Services/BoundarySystem/MixedRealityBoundarySystem.cs
```diff
@@ -624,7 +624,7 @@ public bool Contains(Vector3 location, Boundary.Type boundaryType = Boundary.Typ
             }
 
             // Handle the user teleporting (boundary moves with them).
-            location = MixedRealityToolkit.CameraSystem.CameraRig.PlayspaceTransform.InverseTransformPoint(location);
+            location = MixedRealityToolkit.CameraSystem.MainCameraRig.PlayspaceTransform.InverseTransformPoint(location);
```
StephenHodgson, line 627: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Services/BoundarySystem/MixedRealityBoundarySystem.cs
```diff
@@ -663,7 +663,7 @@ public bool TryGetRectangularBoundsParams(out Vector2 center, out float angle, o
             }
 
             // Handle the user teleporting (boundary moves with them).
-            var transformedCenter = MixedRealityToolkit.CameraSystem.CameraRig.PlayspaceTransform.TransformPoint(
+            var transformedCenter = MixedRealityToolkit.CameraSystem.MainCameraRig.PlayspaceTransform.TransformPoint(
```
StephenHodgson, line 666: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Services/BoundarySystem/MixedRealityBoundarySystem.cs
```diff
@@ -690,7 +690,7 @@ public GameObject GetFloorVisualization()
             }
 
             var floorScale = FloorScale;
-            var position = MixedRealityToolkit.CameraSystem.CameraRig.PlayspaceTransform.position;
+            var position = MixedRealityToolkit.CameraSystem.MainCameraRig.PlayspaceTransform.position;
```
StephenHodgson, line 693: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Services/BoundarySystem/MixedRealityBoundarySystem.cs
```diff
@@ -779,7 +779,7 @@ public GameObject GetTrackedAreaVisualization()
                 layer = DefaultIgnoreRaycastLayer
             };
 
-            var position = MixedRealityToolkit.CameraSystem.CameraRig.PlayspaceTransform.position;
+            var position = MixedRealityToolkit.CameraSystem.MainCameraRig.PlayspaceTransform.position;
```
StephenHodgson, line 782: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Services/DiagnosticsSystem/MixedRealityDiagnosticsSystem.cs
```diff
@@ -124,7 +124,7 @@ public Transform DiagnosticsRoot
                 if (diagnosticsRoot == null)
                 {
                     diagnosticsRoot = new GameObject("Diagnostics").transform;
-                    diagnosticsRoot.parent = MixedRealityToolkit.CameraSystem?.CameraRig.PlayspaceTransform;
+                    diagnosticsRoot.transform.SetParent(MixedRealityToolkit.CameraSystem?.MainCameraRig.PlayspaceTransform, false);
```
StephenHodgson, line 127: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -45,7 +45,7 @@ private GameObject CreateSpatialAwarenessParent
             get
             {
                 var spatialAwarenessSystemObject = new GameObject("Spatial Awareness System");
-                spatialAwarenessSystemObject.transform.parent = MixedRealityToolkit.CameraSystem?.CameraRig.PlayspaceTransform;
+                spatialAwarenessSystemObject.transform.SetParent(MixedRealityToolkit.CameraSystem?.MainCameraRig.PlayspaceTransform, false);
```
StephenHodgson, line 48: We should check if the camera system is null here and fall back to `CameraCache.Main.parent`

XRTK-Core/Packages/com.xrtk.core/Utilities/Solvers/SolverHandler.cs
```diff
@@ -224,10 +224,10 @@ protected virtual void AttachToNewTrackedObject()
                     Handedness = Handedness.Right;
                     break;
                 case TrackedObjectType.Body:
-                    TrackTransform(MixedRealityToolkit.CameraSystem.CameraRig.BodyTransform);
+                    TrackTransform(MixedRealityToolkit.CameraSystem.MainCameraRig.BodyTransform);
```
StephenHodgson, line 227: Not much we can do here. I think solvers should require the usage of the camera system.

PR #518, opened 2020-04-11
Updated Platforms

StephenHodgson:

## Overview

Pulled out some changes from #517 

## Changes:

- Fixed platforms so they only return true outside of the editor
- Fixed a race condition after we've reset the profile and started re-initialization

## Related Submodule Changes

https://github.com/XRTK/Lumin/pull/61


XRTK-Core/Packages/com.xrtk.core/Services/MixedRealityToolkit.cs
```diff
@@ -1799,7 +1799,10 @@ public static IMixedRealityCameraSystem CameraSystem
                     instance.activeProfile == null ||
                     instance.activeProfile != null && !instance.activeProfile.IsCameraSystemEnabled)
                 {
-                    return null;
+                    if (!isResetting && isInitializing)
```
StephenHodgson, line 1802: gonna roll back the changes to the service locator

PR #521, opened 2020-04-13
Dev/platform refactor

StephenHodgson:

## Overview

Pulled out more changes from #517 

I added the ability to only run a service in the editor when only the active build target is available.

## Target of the change:

Is this enhancement for:

## Changes:

Better addresses #378

image
image
image

## Breaking Changes:

- Added `IMixedRealityPlatform.IsBuildTargetAvailable`

## Submodule Changes:

https://github.com/XRTK/SDK/pull/154
https://github.com/XRTK/Lumin/pull/62


FejZa:
Just to be sure. With this change, e.g.

- SimulatedHandControllerDataProvider (Editor, Standalone)
- OculusHandControllerDataProvider (Editor, Android)

only `SimulatedHandControllerDataProvider` will get registered since it's the first in the order to meet the requirement (Editor) and the other one is skipped?

StephenHodgson:
Ah yeah that's not good.

StephenHodgson:
@FejZa I'm open to ideas about how to do this, but It is something I wanna solve before we can finish the Camera and Controller PRs

StephenHodgson:
We don't want to register just the first valid configuration, but we want a way to specify if multiple configurations can be assigned depending on their base types, correct?

FejZa:
I am not sure about how to tackle this. So thinking about the diagnostics system e.g. We have 3 data providers here that are all configured to run "everywhere". And we DO want all of those registered at any time. They all share the base `BaseMixedRealityDiagnosticsDataProvider`.

StephenHodgson:
Yup in this PR all 3 should be running

StephenHodgson:
But for input unrealized that we want to only register one valid configuration per platform, but with a specific constraint of some kind.

I'll think about this some more.

StephenHodgson:
I think I'm going to roll back the changes specifically related to only registering one service per platform.

StephenHodgson:
Okay this PR is ready

XRTK-Core/Packages/com.xrtk.core/Services/MixedRealityToolkit.cs
```diff
@@ -574,7 +574,8 @@ private static void EnsureMixedRealityRequirements()
 
                 availablePlatforms.Add(platformInstance);
 
-                if (platformInstance.IsAvailable)
+                if (platformInstance.IsActive ||
```
FejZa, line 577: I am thinking whether this shouldn't be `&&` instead of `||`? With `||` if my service is configured for Editor, Android it will run as soon as I am in editor, right? But with `&&` we get "it only runs when in editor and the build target is Android". Do we want that? I am unsure.

StephenHodgson: If it's `&&` then the platform would never run in a build.

FejZa: :D true

StephenHodgson:
Okay, I'm feeling much better about this PR.

SimonDarksideJ:
Ultimately, it is down to the developer to define what should be available and when, like with OpenVR, both OpenVR and another platform can run concurrently if you have configured that.
It's up to the developer to decide what should run when.  

So in the simulated case, they could configure:
* Simulated to run on all platforms regardless (activated by a key)
* Simulated could run for the editor, only for specific platforms (e.g. excluding Oculus as hands works natively with Link)
* Simulated may be the only option for a platform (dev / test build 2D without a device)

StephenHodgson:
@SimonDarksideJ I think I covered all these use cases, and even better addressed #378 a bit better.

Pull it down and let me know what you think.

XRTK-Core/Packages/com.xrtk.core/Definitions/Platforms/AllPlatforms.cs
```diff
@@ -9,6 +9,9 @@ namespace XRTK.Definitions.Platforms
     public sealed class AllPlatforms : BasePlatform
     {
         /// <inheritdoc />
-        public override bool IsAvailable => true;
+        public override bool IsActive => true;
```
SimonDarksideJ, line 12: Why have we gone back to "isActive" when we had a whole discussion about changing to "IsAvailable", it's also in contradiction with the "IsBuildTargetAvailable" option.  

Can we revert this... again :D

StephenHodgson: Because it's not active.

Being active is determined by the Service Locator when it gets all the types and resolves the list of `ActiveServices`

StephenHodgson: See https://github.com/XRTK/XRTK-Core/pull/521/files#diff-5ebaa573e8b178b52c80b8c2c7f5873fR580

StephenHodgson: Oh wait I see your point now lol. I'll revert it haha

StephenHodgson: Updated @SimonDarksideJ thanks for the good set of eyes

PR #527, opened 2020-04-15
Add support for input and gestures for hand controllers

FejZa:

## Overview

TODO

## Submodule Changes

- https://github.com/XRTK/SDK/pull/164
- [x] https://github.com/XRTK/WindowsMixedReality/pull/72
- [x] https://github.com/XRTK/Oculus/pull/63
- [x] https://github.com/XRTK/Lumin/pull/68


FejZa:
As the description says this is needed to click buttons etc. / interact with elements in far interaction mode. It's basically "is the mouse button pressed or not".

SimonDarksideJ:
Ok, fair enough will approve.  Although how do you see other gestures (which this seems to be) also be implemented?

FejZa:
It is a gesture, that's right. But we are not doing gesture detection here. This is solely about knowing whether we want to click something. When you say gesture? You mean like navigation / manipulation etc.?

FejZa:
I was hoping we can get those implemented without hand context. Because the gesture is always the same. You click and hold and then move the controller. It's not a hand thing. I might be misunderstanding you though.

SimonDarksideJ:
When I speak of gestures related to hands, I mean those that most hand API's expose, e.g.:

* Open palm
* Fist
* Point
* Pinch
* Squeeze 
* Palm down
* Etc

All of which should raise some sort of gesture event of the type of gesture and pointer information as appropriate.

FejZa:
Maybe we can put it this way. This PR enabels the "Select Gesture" for hands.

SimonDarksideJ:
So do you feel full gesture support for hands is still outstanding then?
I know some API's expose events for them, others rely on "poses" for the hand to recognise them

FejZa:
Ah now I understand. Well this PR adds the "Pinch" gesture if you will. The others are incoming as well in separate PRs.

SimonDarksideJ:
In that case, maybe there needs to be some recognition that a "Pinch" has been detected, else this will need reworking when adding more which will likely follow the same pattern.  Thoughts @StephenHodgson ?

FejZa:
Hmm you're right. I might have jumped the gun here. Let me rework this PR to include all of the mentioned above.

SimonDarksideJ:
The larger question @FejZa is whether Gestures should be assignable to as "Actions", which I suspect may be true.  Each gesture should have an action that can be assigned in the input mapping.

FejZa:
I agree. That was my initial plan. I am just unsure it's currently supported by how the interaction mapping works.

FejZa:
Done

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/Hands/BaseHandController.cs
```diff
@@ -386,8 +407,55 @@ private void UpdateVelocity()
...
+            velocityUpdateFrame = velocityUpdateFrame > velocityUpdateFrameInterval ? 0 : velocityUpdateFrame;
+        }
+
+        private void UpdateIsInputDownPose()
+        {
+            if (TrackingState == TrackingState.Tracked)
+            {
+                var isInInputDownPoseThisFrame = false;
+                if (TryGetJointPose(TrackedHandJoint.ThumbTip, out var thumbTip) &&
+                        TryGetJointPose(TrackedHandJoint.IndexTip, out var indexTip))
```
StephenHodgson, line 420: formatting alignment is off here, but other than that the PR looks good

XRTK-Core/Packages/com.xrtk.core/Providers/Controllers/Hands/BaseHandController.cs
```diff
@@ -84,6 +93,18 @@ public void UpdateController(HandData handData)
                 MixedRealityToolkit.InputSystem?.RaiseSourcePoseChanged(InputSource, this, wristPose);
             }
 
+            if (lastIsInInputDownPose != IsInInputDownPose)
+            {
+                if (IsInInputDownPose && !lastIsInInputDownPose)
+                {
+                    MixedRealityToolkit.InputSystem?.RaiseOnInputDown(InputSource, ControllerHandedness, MixedRealityInputAction.None);
```
StephenHodgson, line 100: I would like to see a way to assign these actions in the controller mapping window somehow.

StephenHodgson:
I noticed we don't raise OnInputClicked

FejZa:
I wasn't sure about `OnInputPressed ` and when to use it. Since you are saying it's continuous. Is it to be raised as soon as the "pointer is down" in this case and then until it's back up again?

StephenHodgson:
That pressed action works for me the way it is.

SimonDarksideJ:
If we look at the way it works in Oculus atm.
OnInputDown begins a countdown to select
Progress bar is filled with OnInputPressed
Selection is activated with OnInputReleasd (if after the progress is full)
Selection is cancelled if OnInputReleased received 

So I believe all three events for gestures would be beneficial going off shell experiences

StephenHodgson:
Might be best to close this PR and make a new branch rebased off of the most recent changes. I'll leave this up to you on how you wanna do it

PR #537, opened 2020-04-23
updated the XRTKs Service locator script execution order to -50

StephenHodgson:

## Overview

On some platforms the services were not setup before some scripts wanted to request services. 

While these scripts should be more resilient to lazy service initialization, we probably do want the locator to run before most everything else in the scene. 

image


SimonDarksideJ:
This change only updates a .Meta file?  Is something missing? @StephenHodgson

StephenHodgson:
That is correct

PR #538, opened 2020-04-23
updated check for current build target

StephenHodgson:

## Overview

Moved the editor check out to the service locator.


XRTK-Core/Packages/com.xrtk.core/Definitions/Platforms/CurrentBuildTargetPlatform.cs
```diff
@@ -22,8 +22,6 @@ public sealed class CurrentBuildTargetPlatform : BasePlatform
         /// <returns>True, if any build target is active.</returns>
         public static bool IsBuildTargetActive(List<IMixedRealityPlatform> platforms)
         {
-            if (!Application.isEditor) { return false; }
-
             var isEditor = false;
```
FejZa, line 25: Almost. There is another "isEditor" validation below. Let's remove that as well for completeness.

StephenHodgson: That's a different check

PR #551, opened 2020-04-28
Concrete Instanced Guid References

StephenHodgson:

## Overview

Previously whenever we refactored any class we were simply serializing the `AssemblyQualifiedName` as a string, which could lead to types being unable to be found after we either rename the class or change the namespace.

This change is backwards compatible to what we had previously.

The two classes to review:

- `SystemType`
- `TypeExtensions`

## Submodule Changes

- https://github.com/XRTK/SDK/pull/166
- https://github.com/XRTK/Lumin/pull/70
- https://github.com/XRTK/Oculus/pull/65
- https://github.com/XRTK/WindowsMixedReality/pull/74

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Editor/PropertyDrawers/PlatformEntryPropertyDrawer.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Editor/PropertyDrawers/PlatformEntryPropertyDrawer.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Runtime.InteropServices;
 using UnityEditor;
 using UnityEngine;
 using XRTK.Definitions;
@@ -18,15 +19,17 @@ namespace XRTK.Editor.PropertyDrawers
         private const string Nothing = "Nothing";
         private const string Everything = "Everything";
         private const string Platform = "Platform";
+        private const string EditorOnly = "Editor Only";
+        private const string EditorAnd = "Editor &";
         private const string TypeReferenceUpdated = "TypeReferenceUpdated";
 
         private static readonly GUIContent TempContent = new GUIContent();
         private static readonly GUIContent EditorContent = new GUIContent("Editor");
         private static readonly GUIContent NothingContent = new GUIContent(Nothing);
         private static readonly GUIContent EverythingContent = new GUIContent(Everything);
-        private static readonly GUIContent EditorOnlyContent = new GUIContent("Editor Only");
+        private static readonly GUIContent EditorOnlyContent = new GUIContent(EditorOnly);
         private static readonly GUIContent RuntimePlatformContent = new GUIContent("Runtime Platforms");
-        private static readonly GUIContent EditorBuildTargetContent = new GUIContent("Editor + Build Target");
+        private static readonly GUIContent EditorBuildTargetContent = new GUIContent($"{EditorAnd} Build Target");
 
         private static readonly int ControlHint = typeof(PlatformEntryPropertyDrawer).GetHashCode();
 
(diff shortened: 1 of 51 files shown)


XRTK-Core/Packages/com.xrtk.core/Editor/PropertyDrawers/PlatformEntryPropertyDrawer.cs
```diff
@@ -313,7 +342,14 @@ void OnEditorSelected(object _)
...
                 var selectedPlatformType = typeRef as Type;
-                var selectedReference = SystemType.GetReference(selectedPlatformType);
+
+                if (selectedPlatformType != null && selectedPlatformType.GUID == Guid.Empty)
+                {
+                    Debug.LogError($"{selectedPlatformType.Name} does not implement a required {nameof(GuidAttribute)}");
+                    return;
+                }
+
+                var selectedReference = selectedPlatformType?.GUID.ToString();
```
SimonDarksideJ, line 352: Can we not make all these references cleaner by having the respective functions simply accept a GUID, then only use "toString" in side the function if needed.  Removing the need for all these "ToString" function calls everywhere.

XRTK-Core/Packages/com.xrtk.core/Runtime/Extensions/TypeExtensions.cs
```diff
@@ -12,22 +15,60 @@ namespace XRTK.Extensions
...
-        /// Attempts to resolve the class reference by the <see cref="Type.AssemblyQualifiedName"/>.
+        /// Attempts to resolve the type using a the <see cref="System.Type.AssemblyQualifiedName"/> or <see cref="System.Type.GUID"/> as <see cref="string"/>.
         /// </summary>
-        /// <param name="classRef">The <see cref="Type.AssemblyQualifiedName"/> of the type to get.</param>
-        /// <returns>The <see cref="Type"/> from <see cref="classRef"/></returns>
-        public static Type ResolveType(string classRef)
+        /// <param name="typeRef">The <see cref="System.Type.GUID"/> or <see cref="System.Type.AssemblyQualifiedName"/> as <see cref="string"/>.</param>
+        /// <param name="resolvedType"></param>
+        /// <returns>The resolved <see cref="Type"/></returns>
+        public static bool TryResolveType(string typeRef, out Type resolvedType)
```
SimonDarksideJ, line 32: Can we also have an overload that takes a GUID as a type instead of a string, so we can avoid string type conversions where we already have a GUID.
(updating this function to also call the GUID version)

StephenHodgson:
> Can we look to provide native GUID functions for areas where we already have GUID references and avoid needless string conversions.

Mainly I'm trying to keep backwards compatibility with any references to `AssemblyQualifiedName`. Doing that expects that all the data is already a guid. Also serializing the `AssemblyQualifiedName` is the fallback in case a class doesn't have the guid attribute.

XRTK-Core/Packages/com.xrtk.core/Runtime/Extensions/TypeExtensions.cs
StephenHodgson: No, because we'd break backwards compatibility with `AssemblyQualifiedName`

XRTK-Core/Packages/com.xrtk.core/Editor/PropertyDrawers/PlatformEntryPropertyDrawer.cs
StephenHodgson: No, because we'd break backwards compatibility with `AssemblyQualifiedName`

XRTK-Core/Packages/com.xrtk.core/Runtime/Extensions/TypeExtensions.cs
StephenHodgson: I'll add the overload

StephenHodgson:
> Also I note you have removed some base constructor references (:Base()) and was curious why

They were redundant

XRTK-Core/Packages/com.xrtk.core/Editor/PropertyDrawers/PlatformEntryPropertyDrawer.cs
SimonDarksideJ: Well do the toString cast inside the function, rather than everywhere that calls the function at the moment, cleaner

StephenHodgson: Some places we're using the serialized property (which may still have the old `AssemblyQualifiedName` serialization.

SimonDarksideJ: Well, once you've created the overload, those places that can should use the new overload, save unnecessary conversion

PR #554, opened 2020-04-29
Added platform profile installer

StephenHodgson:

## Overview

Updated the XRTK profiles so that they live in their respective packages.

**_Default profiles will still be copied on the initial installation of the respective package._**

## Target of the change:

Is this enhancement for:

## Changes:

- fixed some issues with the pathfinder utility
  - relative path needed to add editor/runtime prefixes
- updated symbolic linker to be able to link hidden directories `MyFolder~`
- updated submodules with profile changes
  - removed the platform specific configurations from the SDK profiles
- added `RuntimePlatformAttribute` for `DataProvider` implementations that need to be filtered out in the new `PlatformConfigurationsProfile`

## Submodule Pull Requests

- https://github.com/XRTK/SDK/pull/167
- https://github.com/XRTK/Lumin/pull/71
- https://github.com/XRTK/Oculus/pull/66
- https://github.com/XRTK/WindowsMixedReality/pull/75

## TODO

- [x] Make sure when selecting the service locator object with an empty profile doesn't cause it to ping the object forever.
- [x] Import profiles on first load of package
- [x] Dialog for each platform asking if the default data providers should be added to the registry

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/TypeGrouping.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/TypeGrouping.cs
@@ -14,7 +14,7 @@ namespace XRTK.Definitions.Utilities
         None,
 
         /// <summary>
-        /// No grouping, just show the type names in a lies; for instance, "SpecialClass".
+        /// No grouping, just show the type names in a list; for instance, "SpecialClass".
         /// </summary>
         NoneByNameNoNamespace,
 
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/TypeGrouping.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/TypeGrouping.cs
@@ -14,6 +14,11 @@ namespace XRTK.Definitions.Utilities
         None,
 
         /// <summary>
+        /// No grouping, just show the type names in a lies; for instance, "SpecialClass".
+        /// </summary>
+        NoneByNameNoNamespace,
+
+        /// <summary>
         /// Group classes by namespace and show foldout menus for nested namespaces; for
         /// instance, "Some > Nested > Namespace > SpecialClass".
         /// </summary>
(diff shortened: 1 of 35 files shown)


SimonDarksideJ:
There seem to be a load of cleanup tasks embedded in this PR, can we pull them out?

And what is the plan if the user doesn't hit the copy profiles in the Package Manager? Will it still be available for copy/import when hitting configure?

StephenHodgson:
Of course, it's still in progress.

StephenHodgson:
> And what is the plan if the user doesn't hit the copy profiles in the Package Manager? Will it still be available for copy/import when hitting configure?

Check the TODO

mainly it'll always copy the profiles on first install. But people can pull a fresh copy using the package manager if they want. 

The last step after that is making sure to activate the data providers when a platform package is installed as well.

StephenHodgson:
In the overview I specifically stated :

> Default profiles will still be copied on the initial installation of the respective package.

StephenHodgson:
Just a note when pulling this:

Be sure to reset your default profile generated path to `Assets/XRTK.Generated/` or the symbolic linking of the packages may not set up correctly.

Also be sure that there isn't any profiles in your generated path already (with the exception of the `SymbolicLinkSettings`)

SimonDarksideJ:
Not sure this is the right title for this section on the package manager screen, can it be changed?
image

As the profiles aren't really samples?

Although this also begs the question about adding an "Examples" option later

StephenHodgson:
No it cannot be changed

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/TypeGrouping.cs
```diff
@@ -13,6 +13,11 @@ public enum TypeGrouping
         /// </summary>
         None,
 
+        /// <summary>
+        /// No grouping, just show the type names in a lies; for instance, "SpecialClass".
```
StephenHodgson, line 17: Typo

StephenHodgson, line 17:
```suggestion
        /// No grouping, just show the type names in a list; for instance, "SpecialClass".
```

PR #558, opened 2020-05-01
Dev/better package seed

StephenHodgson:

## Overview

Updated the XRTK.Seed package to list all available packages and installs them all at once.

This also adds compatibility for projects that already contain scoped registres

## Target of the change:

Is this enhancement for:


SimonDarksideJ:
Works well, but why on earth NewtonSoft? (even if only for the few mins the seed exists.)
Also, would it be possible to compile the scripts in to a dll or simply put them in one script called "seed" or "XRTK installeer".  TO simplify the view of the seed package. (not a biggy)

StephenHodgson:
If you can find something else to take its place you're more than welcome to replace it but for now it works.

PR #562, opened 2020-05-04
Added guid regeneration for installed package profiles to prevent asset collisions

StephenHodgson:

## Overview

To prevent asset collisions and problems when copying prefabs and profiles, I thought it'd be best to regenerate any guid references in the copied assets.


SimonDarksideJ:
Does this also solve the issue of asset referencing when the profiles are copied?

StephenHodgson:
Yes, that's the whole point

PR #569, opened 2020-05-05
Update Package installer to work for file extensions

SimonDarksideJ:

## Overview

Resolves issue with package installer which only currently works with .Asset files, extended to now also work with any file extension required by a package.

Also resolves an issue whereby if the destination folder did not exist, it creates it

## Target of the change:

Is this enhancement for:

- Platform (new or updated platform)

## Related changes

 - https://github.com/XRTK/SDK/pull/174


SimonDarksideJ:
Double checking but something is missing there

XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
```diff
@@ -29,11 +29,11 @@ public static class PackageInstaller
         /// <param name="destinationPath">The destination path, typically inside the projects "Assets" directory.</param>
         /// <param name="regenerateGuids">Should the guids for the copied assets be regenerated?</param>
         /// <returns>Returns true if the profiles were successfully copies, installed, and added to the <see cref="MixedRealityToolkitRootProfile"/>.</returns>
-        public static bool TryInstallAssets(string sourcePath, string destinationPath, bool regenerateGuids = true)
+        public static bool TryInstallAssets(string sourcePath, string destinationPath, string extension = "asset", bool regenerateGuids = true)
```
StephenHodgson, line 32: I'd rather just use the search options like in the guid regeneration script and not add the additional parameter.

XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
```diff
@@ -43,9 +43,13 @@ public static bool TryInstallAssets(string sourcePath, string destinationPath, b
                 EditorApplication.delayCall += () => AddConfigurations(installedAssets);
                 return true;
             }
+            else
+            {
+                Directory.CreateDirectory(Path.GetFullPath(destinationPath));
```
StephenHodgson, line 48: I think we should always do this at the top of the method if the directory doesn't exist

StephenHodgson:
Link to ask PR is broken

XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
SimonDarksideJ: It does?  The first line checks if the directory exists.  This is then the "Else" on that check?

XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
SimonDarksideJ: Updated all file functions to use the same method for gathering all known Unity files

XRTK-Core/Packages/com.xrtk.core/Editor/Utilities/UnityFileHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+            "*.anim",
+            "*.prefab",
+            "*.unity",
+            "*.asset",
+            "*.guiskin",
+            "*.fontsettings",
+            "*.controller",
+        };
+
+        public static List<string> GetUnityFiles(string assetsRootPath)
```
StephenHodgson, line 23: Should we also make sure this utility can handle relative/absolute paths?

StephenHodgson: Are we expecting full paths back or relative paths back?

StephenHodgson: Also maybe a better name would be `GetUnityAssetsAtPath` or `GetUnityAssetsInDirectory` and pass in a `DirectoryInfo` instead?

XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
```diff
@@ -119,7 +123,6 @@ private static string CopyAsset(this string rootPath, string sourceAssetPath, st
                 Directory.CreateDirectory(Directory.GetParent(destinationPath).FullName);
 
                 File.Copy(sourceAssetPath, destinationPath);
-                File.Copy($"{sourceAssetPath}{META_SUFFIX}", $"{destinationPath}{META_SUFFIX}");
```
StephenHodgson, line 122: We should def be copying the meta files too.

XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
StephenHodgson: Ah you're right. we have a different path if it exists.

XRTK-Core/Packages/com.xrtk.core/Editor/Utilities/UnityFileHelper.cs
SimonDarksideJ: Everywhere this is used, we only have a string path.  Might consider an overload to use DirectoryInfo, but nowhere would use it at present.  So this is fine.

OK, Will rename to GetUnityAssetsAtPath

XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
SimonDarksideJ: Please read the full scope of the change before commenting.  .meta is one of the file types ALREADY included.  I copied the method used to regenerate GUIDs on your suggestion and that already included Meta files.

Check in the UnityFileHelper that lists all the file types included

SimonDarksideJ:
As stated, the GetUnityAssetsAtPath function now copies the Meta files, so no need to copy them twice!

StephenHodgson:
Nice, I didn't catch that right away. Great work.

Should we also address the guid regeneration after we copied all the files from multiple directories for a package in this PR?

SimonDarksideJ:
Ok, will test and see if we can regen the guids after the folders.

XRTK-Core/Packages/com.xrtk.core/Editor/Utilities/UnityFileHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+            "*.anim",
+            "*.prefab",
+            "*.unity",
+            "*.asset",
+            "*.guiskin",
+            "*.fontsettings",
+            "*.controller",
+        };
+
+        public static List<string> GetUnityAssetsAtPath(string assetsRootPath)
```
StephenHodgson, line 23: Can we add some documentation here?

I'd like to explicitly say that we're returning full paths

SimonDarksideJ: Done

XRTK-Core/Packages/com.xrtk.core/Editor/Utilities/UnityFileHelper.cs
```diff
@@ -8,6 +8,9 @@ namespace XRTK.Editor.Utilities
 {
     public static class UnityFileHelper
     {
+        /// <summary>
+        /// Managed list of recognised Unity files
```
StephenHodgson, line 12:
```suggestion
        /// Managed list of native Unity asset file extensions
```

StephenHodgson:
Now that I think about it, I think if we fail at any point I'd rather just delete any assets we've copied so far.

StephenHodgson:
Don't forget to add the link to the sdk pull request as well please

XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
```diff
@@ -186,5 +187,10 @@ public static void InstallConfiguration(MixedRealityPlatformServiceConfiguration
             AssetDatabase.SaveAssets();
             EditorApplication.delayCall += () => AssetDatabase.Refresh(ImportAssetOptions.ForceUpdate);
         }
+
+        public static void RegenerateGuids(string[] filePaths)
+        {
+            GuidRegenerator.RegenerateGuids(filePaths, false);
```
StephenHodgson, line 193: Would this make more sense to be an overload on `GuidRegenerator` itself?

StephenHodgson: AH I see, I made it internal so people wouldn't use it outside of the core project.

SimonDarksideJ:
Will retest in the morning to see if the new updates don't break the fixes

PR #570, opened 2020-05-06
Feature/service wizard

StephenHodgson:

## Overview

Added menu item to Instance Type dropdown to create a new type based on the `IMixedRealityService` type the picker shows.

A new window will open with the ability to update the output path, namespace, and instance name.

image

image

_Most_ of the class will be implemented for you, with the exception of the nested generic types like `IEqualityComparer` and `IMixedRealityEventSource`. You'll likely need to implement those by easily right clicking on the interface in Visual Studio and selecting `Implement Missing Members`

## Target of the change:

Is this enhancement for:

## Breaking Changes:

To make the wizard work better, I realized that some of our interface types don't follow standard naming conventions and they were adjusted accordingly:

- renamed `IMixedRealitySpatialObserverDataProvider` -> `IMixedRealitySpatialAwarenessDataProvider `
- added `IMixedRealitySystem` to help distinguish services with a single instance


SimonDarksideJ:
A few requests for this change for discussion:

- Can the default folder where scripts are generated emulate the XRTK.Generated pattern.  Have the default be XRTK.UserGenerated. Possibly even have subfolders containing the service code (as some generate multiple) to keep it cleaner
- Can a popup appear just before the services are generated and then popoff when done.  THere was a significant lag when I generated a service (unity froze).  The popup will inform the user it is generating
- Since this is currently generated from a drop down, can the new value of the dropdown be the service just generated?
- For the profile generated, can we also add the template code to generate the profile asset (and assign it?)
- Comments in the templates to give basic guidance for "when to do next" (I can do a PR for that if you wish)

Fantastic work and a bold step forward

StephenHodgson:
> - Can the default folder where scripts are generated emulate the XRTK.Generated pattern.  Have the default be XRTK.UserGenerated. Possibly even have subfolders containing the service code (as some generate multiple) to keep it cleaner

Sure, seems like a trivial change, but I have a feeling that most people will want to put their classes in their own scripts folder, which is why I didn't. But the path also updates the namespace where it goes, so using `XRTK.XXX` would put it in the `XRTK` namespace, which I don't really want to do.

> - Can a popup appear just before the services are generated and then popoff when done.  THere was a significant lag when I generated a service (unity froze).  The popup will inform the user it is generating

The lag is from Unity rebuilding the project. It's already done by this point.

> - Since this is currently generated from a drop down, can the new value of the dropdown be the service just generated?

No, because the class is mostly unimplemented. It would just throw errors if added to the configuration.

> - For the profile generated, can we also add the template code to generate the profile asset (and assign it?)

I don't think this is a good idea either, bc the profile is just empty. I'd rather have someone go and implement their class and profile, then later register it in the configuration profile, and have the configuration profile generate a new asset for them when they press `+`

> - Comments in the templates to give basic guidance for "when to do next" (I can do a PR for that if you wish)

I mean, you implement it, lol. Not sure what else to tell people. lol

FejZa:
I am strongly against generating another root level directory for XRTK. I agree with Stephen, most people want to put their own scripts in their own locations and folders anyways. One thing I liked about XRTK from the beginning compared to MRTK is that it does not add like 5 root directories to my assets folder. 

If needed I'd rather see

XRTK.Generated
- CustomProfiles
- CustomServices

PR #593, opened 2020-05-18
Fix endless configure toolkit dialog issue

FejZa:

## Overview

Fixes an issue where the "There is no active Mixed Reality Toolkit in your scene! Would you like to create one now?" dialog endlessly blocks the editor when pressing "Later".

To observe the error in dev open a scene without the Toolkit in it, e.g. some scene you want to load in additively when running. While in that scene select the MixedRealityToolkitRootProfile asset. The dialog shows up. Press later, press later, press later, press later...

## Changes

Clear selection for one frame if the user answers "Later". That will make sure Unity does not trigger the `OnEnable` event right away again after selectin "Later".

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Editor/Profiles/MixedRealityToolkitRootProfileInspector.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Editor/Profiles/MixedRealityToolkitRootProfileInspector.cs
@@ -51,6 +51,7 @@ namespace XRTK.Editor.Profiles
         private SerializedProperty registeredServiceProvidersProfile;
 
         private MixedRealityToolkitRootProfile rootProfile;
+        private bool didPromptToConfigure = false;
 
         private readonly GUIContent typeLabel = new GUIContent("Instanced Type", "The class type to instantiate at runtime for this system.");
         private readonly GUIContent profileLabel = new GUIContent("Profile");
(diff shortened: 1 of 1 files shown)


FejZa:
It's not fixed it seems. Let me take another look. My first test passed but now it's back sometimes.

FejZa:
Now it's ready

XRTK-Core/Packages/com.xrtk.core/Editor/Profiles/MixedRealityToolkitRootProfileInspector.cs
```diff
@@ -88,7 +89,7 @@ protected override void OnEnable()
             var prefabStage = PrefabStageUtility.GetCurrentPrefabStage();
 
             // Create The MR Manager if none exists.
-            if (!MixedRealityToolkit.IsInitialized && prefabStage == null)
+            if (!MixedRealityToolkit.IsInitialized && prefabStage == null && !didPromptToConfigure)
```
StephenHodgson, line 92: I think having this as an editor pref might be better

FejZa: Hmm not sure how to do that. If I save it to editor prefs "Later" will mean "Never again" unless we implement some logic to clear the value after time x has passed.

FejZa: Nevermind. I can store a timestamp instead of a bool obviously

StephenHodgson: I mean don't reinvent the wheel, we're already doing something similar in `MixedRealityEditorSettings`

StephenHodgson: I think this works as is for now. I think the suggestion was overboard.

PR #616, opened 2020-05-27
Hand input and poses change request

StephenHodgson:

## Overview

Mainly wanted to cleanup the serialization of the recorded joints, and replace the reference types with value types in the data structures. I reserialized all the json structures, and saved them back.

I also fixed the build errors

## Submodule Changes

https://github.com/XRTK/SDK/pull/192

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Controllers/Hands/RecordedHandJoint.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Controllers/Hands/RecordedHandJoint.cs
@@ -20,43 +20,24 @@ namespace XRTK.Definitions.Controllers.Hands
         /// <param name="pose"></param>
         public RecordedHandJoint(TrackedHandJoint joint, MixedRealityPose pose)
         {
-            this.joint = JointNames[(int)joint];
+            this.joint = joint;
             this.pose = pose;
         }
 
-        private static readonly string[] JointNames = Enum.GetNames(typeof(TrackedHandJoint));
+        [SerializeField]
+        private TrackedHandJoint joint;
 
         /// <summary>
-        /// Name of the joint recorded. Rather use <see cref="JointIndex"/> below for convenience.
+        /// Joint this pose represents.
         /// </summary>
-        public string joint;
+        public TrackedHandJoint Joint => joint;
+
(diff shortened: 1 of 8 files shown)


XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Controllers/Hands/RecordedHandJoints.cs
```diff
@@ -10,14 +10,19 @@ namespace XRTK.Definitions.Controllers.Hands
     /// Unity's <see cref="JsonUtility"/> "currently" (stated in 2011) does not support top-level arrays.
     /// But hey, it's on their "future" road-map. That's why we need a wrapper definition around or items.
     /// </summary>
-    /// <remarks>The <see cref="JsonUtility"/> also does not support properties, so we gotta use fields. Oh and it also
-    /// does not support auto mapping of JSON naming "items" to C# naming "Items".</remarks>
```
StephenHodgson, line 14: Actually it does support properties as long as the fields are serialized. Standard naming should be okay in this case. It's when we don't have control of the data structure that we might run into problems here.

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Controllers/Hands/RecordedHandJoint.cs
```diff
@@ -20,43 +20,24 @@ public struct RecordedHandJoint
...
-        /// </summary>
-        public MixedRealityPose pose;
+        [SerializeField]
+        private MixedRealityPose pose;
 
         /// <summary>
-        /// Gets the <see cref="TrackedHandJoint"/> this record represents.
+        /// The recorded pose.
         /// </summary>
-        public TrackedHandJoint JointIndex
```
StephenHodgson, line 42: This seemed a little convoluted and unnecessary so I just replaced it with the `TrackedHandJoint` enum directly.
Enums are pretty good about turning back into their named value, and it didn't seem like we were using the string values but to convert back and forth, so I figured it could go.

FejZa: True, thanks.

PR #636, opened 2020-06-05
Revert API level to 25 for Quest build support

FejZa:

## Overview

Reverts the min API level change in the Core project settings so the project can be built to Oculus Quest.


StephenHodgson:
Been having trouble with the Android builds not finding the sdk lately. Any idea what's causing this?

PR #643, opened 2020-06-16
Removed UNITY_LIGHT_PROBE_PROXY_VOLUME from wireframe shader

StephenHodgson:

## Overview

Sometimes still getting build errors about this shader. I figured it's just best to remove the light probe path anyway, since [unity's built in wireframe shader](https://github.com/TwoTailsGames/Unity-Built-in-Shaders/blob/master/DefaultResourcesExtra/VR/Shaders/SpatialMappingWireframe.shader) is pretty much the same as this one and doesn't use it.


SimonDarksideJ:
Can you test and check this change on ML, as I recall this caused ML to only render in one eye previously.

StephenHodgson:
Magic Leap uses the multiview single pass like all the others now, shouldn't be a problem

StephenHodgson:
> Can you test and check this change on ML, as I recall this caused ML to only render in one eye previously.

I don't ever remember this being an issue. Only for Android?

SimonDarksideJ:
The issue on Android related to the Wireframe shader revolved not rendering at all when an object with the wireframe shader applied was in view.  This was resolved by lowering the graphical quality for Android builds.

However, in our tests, when using just the "EYE_INDEX" properties (after trying to consume the latest MRTK variants) this resulted in only one eye being rendered in ML.

I suggest you run a build using this PR branch on the ML to verify, that's all.

StephenHodgson:
I don't have a device readily available, can you possibly check this?
I honestly don't remember this ever being a problem on Magic Leap. Only Android.

SimonDarksideJ:
OK, I'll whip out my device later.  And honestly, you went all mad because you became a pirate :P

PR #552, opened 2020-04-28
Add support for input and gestures for hand controllers

FejZa:
# XRTK - Mixed Reality Toolkit Change Request

## Overview

This pull request is the next iteration on the hand tracking feature, making it actually useful. It implements a lot of refactoring to improve the overall architecture and data flow based on recent learnings.

### Data Flow / New Platforms

All hand data that platform APIs provide is now being processed in the respective hand data converter. Introduced `BaseHandDataConverter` which contains base and fallback conversion logic for hands, e.g. it will provide a pointer pose if the platform did not provide it. The essential piece of data for hands is now `HandData` more than ever, which the converters generate. So for each platform there is a `PlatformHandControllerDataProvider + PlatformHandDataConverter` pair taking care of preparing data for the agnostic hand controller.

So for a new hand platform "all" that needs to be done is implementing the data provider and if applicable the converter.

### Reduced simulation specifics

The only specifics `SimulatedMixedRealityHandController` now has are the default interaction mappings it needs for the simulated hand to be controlled in editor. It now directly derives from `MixedRealityHandController`. `BaseHandController` got removed since `MixedRealityHandController` essentially is the base for any hand. I'll look into ways to get rid of `SimulatedMixedRealityHandController` completely in a next iteration of hands.

### Near / Far Interaction

Near and far interaction using hands is now supported using
- Pinching gesture (far)
- Poke / Touch gesture (near)

Also added new APIs for
- Pinch strength
- Grab strength

### Physics

The `HandPhysicsEnabled` setting can now be changed at runtime to enable or disable the physics components for hands using the new setter for `IMixedRealityHandControllerDataProvider.HandPhysicsEnabled`. This can be used to e.g. temporarily enable physics interactions and then disable them again. Also physics is now disabled by default.

The `IMixedRealityHandControllerDataProvider.BoundsMode` and `IMixedRealityHandControllerDataProvider.UseTriggers` settings are now runtime settable as well.

### Tracked Poses

All hand profiles now have a new setting "Tracked Poses". These are for one the poses that are available for hand simulation and secondly these are the poses that are being tracked/recognized by the hands system. This is a work in progress feature that will enable user's to define their custom poses they recorded and have input actions raised when those poses are recognized. For now it drives simulation only. Of course each platform may override the global setting here.

tracked_poses

### Rendering Mode

[...]

PR #646, opened 2020-06-19
Fix profiles reset and services reinitialized when already running

FejZa:

## Overview

Fixes the issue reported by @SimonDarksideJ in the hands PR https://github.com/XRTK/XRTK-Core/pull/552

```
Simulated Hands inspector spams console if looked at in runtime

Replication Steps:

Open Demo Scene
Run on any platforms
Select the MRT Game Object
Navigate to "Input System profile"
Expand / touch the IMixedRealityInputDataProvider list
See error SPAM
Error Details

Object reference not set to an instance of an object
  at XRTK.Providers.Controllers.Simulation.BaseSimulatedControllerDataProvider.RefreshSimulatedDevices () [0x00001] in C:\LocalDevelopment\XRTK-Core\XRTK-Core\Packages\com.xrtk.core\Runtime\Providers\Controllers\Simulation\BaseSimulatedControllerDataProvider.cs:129 
  at XRTK.Providers.Controllers.Simulation.BaseSimulatedControllerDataProvider.Update () [0x00008] in C:\LocalDevelopment\XRTK-Core\XRTK-Core\Packages\com.xrtk.core\Runtime\Providers\Controllers\Simulation\BaseSimulatedControllerDataProvider.cs:111 
  at XRTK.Services.MixedRealityToolkit.UpdateAllServices () [0x00095] in C:\LocalDevelopment\XRTK-Core\XRTK-Core\Packages\com.xrtk.core\Runtime\Services\MixedRealityToolkit.cs:1479 
UnityEngine.Debug:LogError(Object)
XRTK.Services.MixedRealityToolkit:UpdateAllServices() (at Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs:1483)
XRTK.Services.MixedRealityToolkit:Update() (at Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs:697)
Expected Behaviour:
No errors while navigating the inspector, whether editor or runtime
```

## Changes

What would happen is that upon selecting / unfolding any data provider configuration in the profiles, while the application is in play mode, causes a full profile reset. This destroys all services and does a full reinitialization of the toolkit. As a side effect of this, the playspace would also get destroyed and recreated. It would then not appear in `DontDestroyOnLoad` collection anymore but in the base scene itself.

To prevent these issues, I disabled resetting profiles when the application is running in editor play mode.

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
@@ -92,6 +92,14 @@ namespace XRTK.Services
         /// <param name="profile"></param>
         public void ResetProfile(MixedRealityToolkitRootProfile profile)
         {
+            if (Application.isEditor && Application.isPlaying)
+            {
+                // The application is running in editor play mode, can't
+                // reset profiles in this state as it will cause destruction
+                // and reinitialization of services in use.
+                return;
+            }
+
             if (isResetting)
             {
                 Debug.LogWarning("Already attempting to reset the root profile!");


FejZa:
Right? I'd like @StephenHodgson to verify this as well as it's his territory and I can't oversee side effects of this change. The question is what happens to changes made to profiles during play mode. I think they are saved to the assets nonetheless but XRTK will not automatically reinitialize when play mode is left. We might want to disable profile editing when in playmode at all.

SimonDarksideJ:
I'm fine merging this.  Unity isn't really meant to be edited at runtime and resets everything back when you finish playing anyway.  So I think this is fine.  If you need to reset in runtime, you still can, just not while in the editor

StephenHodgson:
I guess this is fine for now, but I think it's just a temp fix for whatever you were running into. Can you give me a detailed bug report about it and a commit Sha I can checkout to reproduce it?

SimonDarksideJ:
You mean like the detailed log in this PR and the linked issue comment in the description?

But you don't needs SHA, simply comment out the 4 lines and try the aforementioned replication steps linked

StephenHodgson:
I'd like a Sha commit just cause it was a lot going on.

I didn't see the rest as I was doing a quick reply from the app. Didn't see the rest of the thread. Sorry.

I'll do my best to check it out when I have a chance. Just seems like a cover up of an issue that could have been fixed a proper way. Again I think it's fine for now but we do want the ability to reset profiles at runtime later as we add more features

FejZa:
It's a valid fix I think as I don't see any use case and scenario where you'd want a full reset of the service locator and all services etc. while the application is running. Also this affects only in-editor work. You can still do anything as before in a built player. This is only about fiddling with profiles in the inspector while the game is running.

StephenHodgson:
I'd still like to understand what was happening and why it wasn't a problem before now.

FejZa:
I'll post a recording of what is going on over on Discord

StephenHodgson:
After reviewing this should be fine. I agree with Dino's original assessment. We probably shouldn't be resetting the service locator in playmode while in the editor

Issue #656, opened 2020-07-25
DefaultHeadHeight setting in camera data provider profile causes misbehaviour


FejZa:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

The `DefaultHeadHeight` setting is causing issues in my WMR hand tracking implementation. Upon research it seems to me, the setting is not doing what it's supposed to do or it's unclear to me what problem it is trying to solve.

**What I understand it should do**
The default head height should only be applied, if the target device does NOT provide head height itself, e.g. Oculus GO or other 3DoF devices. So the setting should NOT affect HoloLens 2 as I understand.

**What it does it seems**
The setting gets applied as soon as the target device is stereoscopic, no matter whether it is 6DoF or not and provides head pose natively.

The properties docs say
```
/// <summary>
/// The default head height when a platform doesn't automatically set it.
/// </summary>
float DefaultHeadHeight { get; }
```

So it should only do anything `when a platform doesn't automatically set it`, which HoloLens 2 does. So I figure this is a bug.

**What I observe**

I set the default head height to `1.6`, following happens
- Run application while seated -> Camera Y-Position = 1.6 (this does not make sense as I am not 1.6 above floor when seated)
- Now stand up -> Camera Y-Position ~= 2.2 (this does not make sense as I am not 2.2m tall)

## To Reproduce

- Set the default head height to e.g. 1.6 and run either on device or using holographic remoting
- Observe the camera transform position as you move around

## Expected behavior

- The setting is ignored on HoloLens 2 and any other 6DoF device.

## Actual behavior

- The setting is not ignored

PR #657, opened 2020-07-30
Only apply default head height if not managed by device

FejZa:

## Overview

Introduced 

```
/// <summary>
/// Is the head height, and thus the camera y-position, managed by the device itself?
/// If true, the <see cref="DefaultHeadHeight"/> setting is ignored and has no effect
/// on camera positioning.
/// </summary>
bool HeadHeightIsManagedByDevice { get; }
```

to `IMixedRealityCameraDataProvider` and implemented as

```
/// <inheritdoc />
public virtual bool HeadHeightIsManagedByDevice => XRDevice.isPresent;
```

in `BaseCameraDataProvider`

to only apply the `DefaultHeadHeight` setting, if the device does not manage head height itself, which is the case
with many devices, like HoloLens e.g.

By default we will assume as soon as an XR Device is connected that it manages head height, since most devices these days will do. The developer has the option to override the default value per platform.

## Changes

- Fixes: #656

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Interfaces/CameraSystem/IMixedRealityCameraDataProvider.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Interfaces/CameraSystem/IMixedRealityCameraDataProvider.cs
@@ -19,6 +19,13 @@ namespace XRTK.Interfaces.CameraSystem
         bool IsStereoscopic { get; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


FejZa:
@SimonDarksideJ I tested this and it did resolve the issues I had. Ready for review.

FejZa:
True, I'll double check the platform overrides.

SimonDarksideJ (approved):
Approved, tested on WMR, Oculus and OpenVR.  Hands now appear correctly on all these platforms.

PR #693, opened 2020-11-24
Fixed min attribute. now properly clamps min value in inspector

StephenHodgson:

## Overview

- added checks for near clips values in the camera data provider to prevent them from being set to zero
- updated all submodules

## Changes

- Fixes: https://github.com/XRTK/WindowsMixedReality/issues/93


FejZa (approved):

FejZa:
PR looks good. Personally I don't see a scenario where profile properties need a Setter. Profile are just for in-editor configuration using the inspector. People shouldn't change profile values in code.

PR #695, opened 2020-11-24
Fix NRE on package asset import

FejZa:

## Overview

Addresses https://github.com/XRTK/Oculus/issues/92

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
@@ -139,20 +139,33 @@ namespace XRTK.Editor
 
                 if (platformConfigurationProfile.IsNull()) { continue; }
 
-                var rootProfile = MixedRealityToolkit.IsInitialized
-                    ? MixedRealityToolkit.Instance.ActiveProfile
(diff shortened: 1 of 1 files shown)


XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
```diff
@@ -139,20 +139,33 @@ private static void AddConfigurations(List<string> profiles)
...
+                MixedRealityToolkitRootProfile rootProfile;
+                if (MixedRealityToolkit.IsInitialized)
                 {
-                    InstallConfiguration(platformConfigurationProfile, rootProfile);
+                    rootProfile = MixedRealityToolkit.Instance.ActiveProfile;
                 }
                 else
                 {
-                    EditorUtility.DisplayDialog("Attention!", "Each data provider will need to be manually registered in each service configuration.", "OK");
+                    var availableRootProfiles = ScriptableObjectExtensions.GetAllInstances<MixedRealityToolkitRootProfile>();
```
StephenHodgson, line 149: Do we also need to make this change in all the other package installers?

FejZa: I don't think so. The other platform package installers are using this one in core. The code is shared.

SimonDarksideJ (approved):

PR #704, opened 2020-11-27
updated playmode/editor mode change transitions

StephenHodgson:

## Overview

- Disable and destroy all services as soon as we get the quitting event
- Properly use `nameof`
- Fixed manifest

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
@@ -289,8 +289,8 @@ namespace XRTK.Services
                 {
                     if (instance != null)
                     {
-                        Debug.Assert(instance.transform.parent == null, "The MixedRealityToolkit should not be parented under any other GameObject!");
-                        Debug.Assert(instance.transform.childCount == 0, "The MixedRealityToolkit should not have GameObject children!");
+                        Debug.Assert(instance.transform.parent == null, $"The {nameof(MixedRealityToolkit)} should not be parented under any other GameObject!");
+                        Debug.Assert(instance.transform.childCount == 0, $"The {nameof(MixedRealityToolkit)} should not have GameObject children!");
                     }
                 }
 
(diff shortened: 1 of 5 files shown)


SimonDarksideJ (dismissed):
Makes sense, testing will tell :D

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
```diff
@@ -274,7 +274,12 @@ private void InitializeInstance()
                     DontDestroyOnLoad(instance.transform.root);
                 }
 
-                Application.quitting += () => IsApplicationQuitting = true;
+                Application.quitting += () =>
```
FejZa, line 277: I wouldn't have used a lambda here but I guess it's too late.

StephenHodgson: Why not?

FejZa: Just a personal preference. Application exit and cleanup is an important event and imo deserves a proper method definition. That makes it for others easier to find the code executed when the toolkit is "deinitialized" in the future, instead of having to find it somewhere in between other code.

But as I said, it's not super important. I just wouldn't have done it.

StephenHodgson: It's an event you can subscribe to anywhere. We do provide this public flag for if the application is quitting via the property, but I don't really think it makes a difference if the `Application.quitting` event can be subscribed from anywhere in the codebase

FejZa: I think you misunderstood. All I am saying is I'd create a definition `private void MixedRealityToolkit_OnQuitting()` in `MixedRealityToolkit` instead of an anonymous method. Because that way if someone is looking up "What does the XRTK do, when the application quits?", it is much easier to find. All good.

StephenHodgson: Now that makes sense !👍

Issue #706, opened 2020-11-29
SDK references are lost when a Specific Platform has a dependency 


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

Currently, SDK assets are copied to the user's project when installed, this causes the GUID's for those assets to be regenerated to avoid conflicts.

However, if a platform's configuration has a dependency on SDK assets, the references they had stored will no longer be correct, resulting in bad configuration.

## To Reproduce

1. Create a new XRTK project
2. Register the Scoped Registry and Install the XRTK.SDK package (which automatically registers the Core)
3. Wait for package install, copy and guid regen to complete
4. Add another platform, e.g. Oculus
5. When prompted, say "Yes, Absolutely" to register the service configuration automatically
6. Wait for platform to install
7. Run "Mixed Reality Toolkit -> Configure" to setup the scene and initial configuration
8. Select the MixedRealityToolkit Gameobject and observe the configuration screen in the inspector
9. Navigate to "XRTkRoot -> Input System Profile -> Oculus Controller Data Provider profile -> Advanced -> Lect Oculus Touch profile -> Spatial Pointer Profile"
10. Observe Registered Pointer Profiles are null (SDK references lost)

This occurs anywhere there is a reference to an SDK component in a platform

## Expected behavior

The SDK reference should be preserved and the configuration should be valid

## Actual behavior

Broken "Null" references
(also observed in Platform Hand Pointer configuration, with a reference to the SDK Hand Pointer Default Pointer prefab)

## Your Setup

- Operation system Windows Standalone / Android
- Unity Version 2019.4.15
- XRTK Version 0.2.0-preview

## Additional context

Discussed in #Troubleshooting channel on Discord

StephenHodgson:
I suppose the best we can do is create a map from the known guids from the sdk to the new regenerated ones, and so a search/replace on them.

Does this still happen if we install the platforms _first_ then install the SDK?

SimonDarksideJ:
Happens either way, simple fact, the GUID's in the UPM package are different to the GUID's when copied to the asset folder, so yet, they will always be broken

SimonDarksideJ:
Tested with UPM, sadly the issue failed testing

StephenHodgson:
How did it fail?

StephenHodgson:
I had forgotten to update the sdk with the breaking changes. The PR is open for that now.

StephenHodgson:
[...]

PR #709, opened 2020-12-10
Included platform prefabs in the Symbolic links for dev branch

SimonDarksideJ:

## Overview

Included platform prefabs in the Symbolic links for dev branch

## Changes

Symbolic links updated to include new package hidden folders

## Related changes

- https://github.com/XRTK/Oculus/pull/95
- https://github.com/XRTK/WindowsMixedReality/pull/108


StephenHodgson (changes requested):
There shouldn't be prefabs in the platform packages

SimonDarksideJ:
There should not be dependencies between the Platforms and another package, e.g. the SDK.  This breaks platforms.
We can work on a better solution for 0.3

StephenHodgson:
I'd much rather see a dependency on the SDK then having to manage duplicate assets across all the platforms

SimonDarksideJ:
If you can resolve #706 in short order, we can reevaluate the way forward. Else we will need to accept this approach in the short term to lead to a future workable arrangement

Issue #504, opened 2020-04-06
Teleporting does not land on desired landing location


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

Currently, when a user teleports to a location, their landing spot is not where they selected, for example

![image](https://user-images.githubusercontent.com/1793042/78596630-0843ed00-7844-11ea-932a-7b0e2668b0cd.png)

1. User in left area selects a teleport location
2. User lands in a space "relative" to where they selected, as shown on right (different based on teleport direction)

GIF of this happening in the editor
![Development-TeleportationBroken-2](https://user-images.githubusercontent.com/1793042/78596502-d03caa00-7843-11ea-9aef-c820ee117635.gif)

## To Reproduce

1. Create project and configure
2. Add floor to teleport on
3. Teleport to an obvious location, e.g. Corner
4. Observe landing location

Additional replication

1. Create project and configure
2. Add floor to teleport on
3. Add a new plane, size to a small square and apply a red material
4. Add "Teleport Hostspot" SDK component
5. Teleport to HotSpot
4. Observe landing location

## Expected behavior

Teleport should land where they selected

## Additional notes
Teleport system is a mess :D  and needs an overhaul.  Could not assign the teleport activity to a button.  Post 0.2 change to teleport system.

StephenHodgson:
What platform were you testing this out on?

SimonDarksideJ:
ALL of them. Primarily WMR, Oculus and OpenVR

StephenHodgson:
Works correctly on Lumin last I checked. I wonder if it's updating the position even though you've let go and continued to move your hand?

StephenHodgson:
I did make some changes in #517 to the Teleport system to ensure we're at least using the [correct height](https://github.com/XRTK/XRTK-Core/pull/517/files?file-filters%5B%5D=.cs#diff-b63a3759cb94eba085c6d2e4e4a0d699R224)

SimonDarksideJ:
There is a possibility that the movement is picking up / changing based on the letting go of the thumbstick, but no matter how careful I am, it does.
Might try changing the teleport action to a button to verify.

But ultimately, this shouldn't happen, the place the pointer is when you let go should be the location you land.

StephenHodgson:
Yeah I think some verification tests should do the trick. 

I was thinking about putting a small sphere in the place the user lets go and the pointer position recorded. Then once the teleport is finished looking down to see how close you came.

[...]

PR #722, opened 2020-12-17
Teleport System v2

FejZa:

## Overview

This PR fixes https://github.com/XRTK/XRTK-Core/issues/504 and introduces some new features to the teleport system.
The core change is that the act of "teleporting" is now done by "teleportation handlers". Instead of the teleport system just moving the camera around. Using a new teleport system profile the user can configure a handler implementation to use, this opens up endless possibilities to handle teleportation.

Previously:
- Teleport just moves the camera/playspace to target position instantly. That's it

Now:
- You can either just teleport or have e.g. an animation / sound or whatever play when teleporting. It's all up to how the handler implements it

The PR comes with two default implementations.

### `InstantTeleportHandler`

This is what we had so far in terms of looks and behaviour. The user selects a teleport location and instantly is moved there.

instant-teleport_1

### `FadingTeleportHandler`

This implementation will fade in / out the camera before teleporting, which leads to a more appealing and smooth teleport.

fade-teleport_1

## Changes

- Fixes: https://github.com/XRTK/XRTK-Core/issues/504

## Submodule Changes

- [SDK](https://github.com/XRTK/SDK/pull/213)
- [Examples](https://github.com/XRTK/Examples/pull/19)

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Services/TeleportSystem/MixedRealityTeleportSystem.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Services/TeleportSystem/MixedRealityTeleportSystem.cs
@@ -30,8 +30,6 @@ namespace XRTK.Services.Teleportation
         private TeleportEventData teleportEventData;
         private bool isTeleporting = false;
         private bool isProcessingTeleportRequest = false;
-        private Vector3 targetPosition = Vector3.zero;
-        private Vector3 targetRotation = Vector3.zero;
 
         #region IMixedRealityService Implementation
 
@@ -196,14 +194,13 @@ namespace XRTK.Services.Teleportation
         {
             isProcessingTeleportRequest = true;
 
-            targetRotation = Vector3.zero;
-            targetPosition = eventData.Pointer.Result.EndPoint;
+            var targetRotation = Vector3.zero;
+            var targetPosition = eventData.Pointer.Result.EndPoint;
             targetRotation.y = eventData.Pointer.PointerOrientation;
 
             if (eventData.HotSpot != null)
             {
                 targetPosition = eventData.HotSpot.Position;
-
                 if (eventData.HotSpot.OverrideTargetOrientation)
                 {
                     targetRotation.y = eventData.HotSpot.TargetOrientation;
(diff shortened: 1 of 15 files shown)


XRTK-Core/Packages/com.xrtk.core/Runtime/Services/TeleportSystem/MixedRealityTeleportSystem.cs
```diff
@@ -190,45 +196,10 @@ public void RaiseTeleportCanceled(IMixedRealityPointer pointer, IMixedRealityTel
...
-            var height = targetPosition.y + parentPosition.y;
-            targetPosition -= cameraPosition - parentPosition;
-            targetPosition.y = height;
-            cameraParent.position = targetPosition;
-            cameraParent.RotateAround(cameraPosition, Vector3.up, targetRotation.y - cameraTransform.eulerAngles.y);
-
-            isProcessingTeleportRequest = false;
-
-            // Raise complete event using the pointer and hot spot provided.
-            RaiseTeleportComplete(eventData.Pointer, eventData.HotSpot);
```
StephenHodgson, line 231: I take this all this was moved to the TeleportHandler implementation?

StephenHodgson:
Overall love the work, I think it's great we have a proper profile and handlers for this now.

I guess my only concern is that the default implementation cannot be packaged with the core (similar to the gaze data provider which serves a similar function as it is required for the input system to work)

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/TeleportSystem/MixedRealityTeleportSystem.cs
FejZa: Right

FejZa:
But don't most of the systems have that issue? Spatial Awareness also won't be able to do anything without SDK.

FejZa:
We could have a "fallback default implementation" that kicks in if no handlers is available. As soon as one is registered in the system profile it takes over?

StephenHodgson:
Yeah but the difference with the Spatial Awareness system is that not all platforms could have data providers.

I think it's more like the default gaze provider.

StephenHodgson (approved):
Got to try out the demos, great stuff Dino!

PR #752, opened 2021-01-12
Fix auto scene switcher

FejZa:

## Overview

This PR fixes the `AutoSceneSwitcher` showing up when the toolkit is actually already configured in the scene(s).
Also this fixes the `AutoSceneSwitcher` forcing the XRTK base scene to be the active scene. This breaks multi-scene editing behaviour expected from the editor. As only the active scene's lightmaps will be used and people might want another scene to be the active scene in editor. The XRTK does not require the base scene to be the active scene, it just needs to be loaded.

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Editor/Utilities/AutoSceneSwitcher.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Editor/Utilities/AutoSceneSwitcher.cs
@@ -1,10 +1,12 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) XRTK. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using UnityEditor;
 using UnityEditor.Experimental.SceneManagement;
 using UnityEditor.SceneManagement;
 using UnityEngine.SceneManagement;
+using XRTK.Services;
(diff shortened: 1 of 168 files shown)


StephenHodgson:
This is the intended behavior

StephenHodgson:
You should be able to open and edit multiple scenes at once in the editor

FejZa:
What is intended behaviour?

FejZa:
I can open multiple scened and edit them yes, that is not the issue here.

SimonDarksideJ:
I've hit this a few time before which has led me not to use the base scene for these very issues.  Seeing as this fix resolves them, I would welcome it.

SimonDarksideJ (dismissed):

StephenHodgson:
This passed already and I cancelled it bc it will loop forever

FejZa:
I see. I am trying to understand what's wrong with the docs build.

StephenHodgson:
@FejZa could you cherry pick the changes from this PR and open a new one so the doc changes are left behind?

FejZa:
WTH happened here. This PR was a single file changed and no triple slash docs changed at all. How come I have 169 changed files now.

StephenHodgson:
The docs build step. I removed that step now.

FejZa:
Closed in favor of https://github.com/XRTK/XRTK-Core/pull/754

Issue #758, opened 2021-01-15
XRTK auto-configures without input from user


StephenHodgson:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

Calling `Instance` on the xrtk causes the system to auto configure without user input.
https://github.com/XRTK/XRTK-Core/blob/1494d3f8c7bcb738923cfbd140a4639b4864ddea/XRTK-Core/Packages/com.xrtk.core/Editor/EditorActiveProfileChangeHandler.cs#L21

## To Reproduce

1. Open any of the submodule projects
2. Create a new scene
3. XRTK prompts you for a profile
4. See the scene configure itself without input from you, the user.

## Expected behavior

User should explicitly go to the mixed reality toolkit menu to configure the xrtk

## Actual behavior

When opening a project or new scene the xrtk tried to configure by itself.

## Your Setup

- XRTK Version 0.2.1

StephenHodgson:
Seems as though this is still happening. 

Steps to test:

1. Create a new scene
2. Ensure no start scene is setup in the Preferences/XRTK
3. Enter playmode

PR #760, opened 2021-01-15
Fixed auto-configuration without user input

StephenHodgson:

## Overview

fixed a bug where the xrtk would auto-configure itself bc the Instance was called

## Changes

- Fixes: #758


FejZa (approved):
Nice! Thanks for looking at this.

PR #766, opened 2021-01-22
Implement utility to check for used render pipeline

FejZa:

## Overview

Adds a small utility class to detect the render pipeline currently in use.

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/RenderPipeline.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/RenderPipeline.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) XRTK. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace XRTK.Definitions.Utilities
+{
+    /// <summary>
+    /// Available render pipelines for use in Unity applications.
+    /// </summary>
+    public enum RenderPipeline
+    {
+        /// <summary>
+        /// The legacy "built-in" render pipeline of Unity that was deprecated
+        /// when scriptable render pipelines were introduced.
+        /// </summary>
+        Legacy = 0,
+        /// <summary>
+        /// The universal render pipeline, formerly also known as lightweight render pipeline.
+        /// </summary>
+        UniversalRenderPipeline,
+        /// <summary>
+        /// The high definition pipeline.
+        /// </summary>
+        HighDefinitionRenderPipeline,
+        /// <summary>
+        /// A customized scriptable render pipeline.
+        /// </summary>
+        Custom
+    }
+}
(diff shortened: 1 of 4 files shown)


XRTK-Core/Packages/com.xrtk.core/Runtime/Utilities/RenderPipelineUtilities.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            var renderPipelineAsset = GraphicsSettings.renderPipelineAsset;
+            if (renderPipelineAsset.IsNull())
+            {
+                return RenderPipeline.Legacy;
+            }
+
+            switch (renderPipelineAsset.GetType().Name)
+            {
+                case urpAssetTypeName:
```
StephenHodgson, line 29: Wondering if we can use pattern matching for the types instead of using the string name

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/RenderPipeline.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) XRTK. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace XRTK.Definitions.Utilities
+{
+    /// <summary>
+    /// Available render pipelines for use in Unity applications.
+    /// </summary>
+    public enum RenderPipeline
```
StephenHodgson, line 9: Unity doesn't have an enum like this?

StephenHodgson (commented):
See comments

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/Utilities/RenderPipeline.cs
FejZa: Nope, unfortunately not.

XRTK-Core/Packages/com.xrtk.core/Runtime/Utilities/RenderPipelineUtilities.cs
FejZa: The actual types are only available if the URP / HDRP package is installed and this check needs to work even when no render pipeline package is in use. So string is all we got.

StephenHodgson: @FejZa do you know if those packages define symbols for the features?

FejZa: No they don't. Freya Holmér put a lot of thoughts into this and this is pretty much what she came up with in the end. It's not perfect but I don't think currently there is any better way to check this.

SimonDarksideJ (approved):
Looks good to me. Did a bit of poking around and agree, it's the simplest solution since Unity doesn't provide an internal API

PR #791, opened 2021-02-14
Delay showing the dialog until the next editor gui update

StephenHodgson:

## Overview

Delay showing the dialog for no profile when setting up to prevent dialog cancellation on mac

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Editor/MixedRealityToolkitInspector.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Editor/MixedRealityToolkitInspector.cs
@@ -76,7 +76,8 @@ namespace XRTK.Editor
                 {
                     case 0:
                         EditorGUIUtility.PingObject(target);
-                        EditorUtility.DisplayDialog("Attention!", "No root profile for the Mixed Reality Toolkit was found.\n\nYou'll need to create a new one.", "OK");
+                        EditorApplication.delayCall += () =>
+                            EditorUtility.DisplayDialog("Attention!", "No root profile for the Mixed Reality Toolkit was found.\n\nYou'll need to create a new one.", "OK");
                         break;
                     case 1:
                         var rootProfilePath = AssetDatabase.GetAssetPath(rootProfiles[0]);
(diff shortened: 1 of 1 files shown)


XRTK-Core/Packages/com.xrtk.core/Editor/MixedRealityToolkitInspector.cs
```diff
@@ -76,7 +76,8 @@ public override void OnInspectorGUI()
                 {
                     case 0:
                         EditorGUIUtility.PingObject(target);
-                        EditorUtility.DisplayDialog("Attention!", "No root profile for the Mixed Reality Toolkit was found.\n\nYou'll need to create a new one.", "OK");
+                        EditorApplication.delayCall += () =>
```
FejZa, line 79: I am not sure with these .delayCall registration. Same thing as with the package installer we talked about yesterday. Pretty sure this is causing memory leaks in the editor.

StephenHodgson: They're just closures. Should be okay.

It's a lot harder to create memory leaks I'm c# unless you're playing with low level memory

StephenHodgson: @FejZa can you find me an example usage of this delay call that doesn't use an anonymous delegate?

FejZa: Like in XRTK? E.g. here https://github.com/XRTK/WindowsMixedReality/blob/development/XRTK.WindowsMixedReality/Packages/com.xrtk.wmr/Editor/WindowsMixedRealityPackageInstaller.cs

FejZa: I don't know. To me this looks like every GUI frame we are registering an event handler. But I might be wrong here. I just live by the rule, if I subscribe, I unsubscribe.

FejZa (approved):

XRTK-Core/Packages/com.xrtk.core/Editor/MixedRealityToolkitInspector.cs
StephenHodgson: No just in general.

StephenHodgson: Let's take this discussion to discord :)

FejZa:
Well, should have read the docs:

"Delegate which is called once after all inspectors update.
Add functions to this delegate in order to delay their execution until after inspectors have updated. Each function is only executed once after it is added."

It seems Unity removes every registered method once called automatically.

StephenHodgson:
> It seems Unity removes every registered method once called automatically.

That was my understanding as well

StephenHodgson:
image

Issue #800, opened 2021-03-11
Remove current gLTF utility


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Feature Request

## Is your feature request related to a problem? Please describe

Very simple, remove the current gLTF utilities from CORE.  They cause more issues than they solve.

Additionally, because it is part of Core, it makes it impossible for an XRTK user to use an alternative.

If we still want to provide our version, then it should be extracted into its own project and updated with the latest capabilities, etc

FejZa:
"Additionally, because it is part of Core, it makes it impossible for an XRTK user to use an alternative."

I can see this being an issue and removing them makes sense.

You mention other issues. What are those?

StephenHodgson:
Isn't this a duplicate?

I was gonna try to get around this pretty.

Afaik there's no real bickers besides the build window, but we could just make the gltf asset a prefab reference I suppose.

SimonDarksideJ:
I think that issue @StephenHodgson should be moved to the separate project should we wish to continue it.

As to issues @FejZa , I have utterly failed to get gLTF models to load now and it complains about gLTF buffers being null.  Granted, these are streamed gLTF's, but I already had this working, but now (when I tested again this week) it utterly fails.

But primarily because this is in core, I can't use another solution without drastic measures.

PR #801, opened 2021-03-11
Extract glTF utility to its own project.

SimonDarksideJ:

## Overview
PR to extract the glTF utilities in to their own project.

Having a specific gLTF library within the XRTK restricted used from using their own glTF libraries and binds XRTK to tightly on a single glTF solution.

This PR simple removes the current glTF solution and moves it to it's own optional repo

## Changes

- Fixes: #800

## Breaking Changes

- Breaks
- Only breaking change is that this removes the native ability to load glTF controller models directly.  This feature should be refactored to be more open.

## Related Submodule Changes

- https://github.com/XRTK/Examples/pull/31
- https://github.com/XRTK/glTF/tree/development
- https://github.com/XRTK/WindowsMixedReality/pull/126


XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/Controllers/BaseController.cs
```diff
@@ -252,23 +233,10 @@ public async Task TryRenderControllerModelAsync(byte[] glbData = null, bool useA
                     ? cameraSystem.MainCameraRig.PlayspaceTransform
                     : CameraCache.Main.transform.parent;
 
-                // If the model was loaded from a system template
-                if (gltfObject != null)
```
StephenHodgson, line 256: We need to provide an alternative

StephenHodgson (changes requested):
See comment

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/Controllers/BaseController.cs
SimonDarksideJ: The alternative is ultimately to use models in your project obviously.

Another way is to have in the sdk have a feature that can generate the model and then set the model parameter using a gender gltf interface.

But without a gltf system built into the core, the core can't enable that. But tbh most projects don't use this

StephenHodgson: I'm using the gltf to load controller models for wmr and oculus

SimonDarksideJ: So what is your suggestion to address this if you are loading models?  Given the aim is to allow users to use their preferred glTF importer.

StephenHodgson: Can we at least provide a game object to instantiate?

FejZa:
My question would be, why does XRTK needs its own implementation of glTF / its own glTF module? This is something I haven't been using at all yet and thus can't really judge.

So why do we need this and what is the resolution for this PR?
@StephenHodgson  "Can we at least provide a game object to instantiate?" can you explain what you mean? We are already creating a game object for each controller aren't we?

SimonDarksideJ:
> My question would be, why does XRTK needs its own implementation of glTF / its own glTF module? This is something I haven't been using at all yet and thus can't really judge.

The separate implementation is simply another variant on the Khronos implementation but updated to use Unity JSON and a host of other improvements.  I think it's worth preserving and maintaining in its own repo.  But being separate means you can use your own if you wish instead.

> So why do we need this and what is the resolution for this PR?
> @StephenHodgson "Can we at least provide a game object to instantiate?" can you explain what you mean? We are already creating a game object for each controller aren't we?

That's what confused me, just hadn't had the time to circle back to this conversation.  Any further thoughts @StephenHodgson ?

StephenHodgson:
Yes this is true, for edit time, but if we have a 3rd party glTF importer it'll likely pass a GameObject reference as either the instantiated object or an object to insatiate. I suppose that's what I'm looking for exactly. Just to link the references up (if the object has already been created) or to create it for us if it's not already in the scene.

FejZa:
Fair enough, now I get it.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/Controllers/BaseController.cs
SimonDarksideJ: OK, doing some testing and the simplest way would be for the developer to update the visualizer type for the controller to be one to load the glTF model required.

You don't need a GameObject to attach to, simply use your own visualizer that loads your glTF model.

Thoughts @StephenHodgson

SimonDarksideJ:
> but if we have a 3rd party glTF importer it'll likely pass a GameObject reference

@StephenHodgson  As commented above, you don't need a GameObject reference, just write your own Visualizer to load the glTF model and set the Visualizer type to your new visualiser for the controllers you want to configure.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/Controllers/BaseController.cs
StephenHodgson: Where is `gltfObject` declared? Shouldn't we remove that as well?

SimonDarksideJ: > Where is `gltfObject` declared? Shouldn't we remove that as well?

Line 222, it was removed.  THere is no longer any glTF reference in the Core classes post this change

StephenHodgson:
What is failing the build?

XRTK-Core/Packages/com.xrtk.core/Editor/BuildAndDeploy/BuildDeployWindow.cs
```diff
@@ -525,16 +524,16 @@ private void AppxBuildGUI()
 
             var prevFieldWidth = EditorGUIUtility.fieldWidth;
 
-            EditorGUIUtility.fieldWidth = 120;
-            var appIconAssetObject = AssetDatabase.LoadAssetAtPath<GltfAsset>(UwpBuildDeployPreferences.MixedRealityAppIconPath);
+            //EditorGUIUtility.fieldWidth = 120;
+            //var appIconAssetObject = AssetDatabase.LoadAssetAtPath<GltfAsset>(UwpBuildDeployPreferences.MixedRealityAppIconPath);
```
StephenHodgson, line 529: Let's just go ahead and remove the lines altogether. I'm rewriting this anyway and don't want conflicts.

SimonDarksideJ: Darn, I thought I had deleted everything.  Good spot.

SimonDarksideJ:
> What is failing the build?

Just checked, it's complaining about the Examples project.  I'll do a quick PR to also remove the glTF samples from there

SimonDarksideJ:
Pushed an update to change the branch for examples and the recent comments

SimonDarksideJ:
@StephenHodgson Issues resolved

StephenHodgson (changes requested):
Let's keep the glTF examples and just update the references

StephenHodgson:
@SimonDarksideJ let's be sure to create a issue in WMR for adding the visualizer later on once the glTF package is published.

StephenHodgson (approved):

PR #810, opened 2021-04-15
removed the min unity release rule

StephenHodgson:

## Overview

Removed the minimum unity release version for all XRTK projects. Any version of Unity 2019.4.X should be compatible.

### DIFF
--- a/XRTK-Core/Packages/manifest.json
+++ b/XRTK-Core/Packages/manifest.json
@@ -15,8 +15,6 @@
     "com.unity.ide.vscode": "1.2.3",
     "com.unity.mobile.android-logcat": "1.2.1",
     "com.unity.test-framework": "1.1.24",
-    "com.unity.textmeshpro": "2.1.4",
-    "com.unity.ugui": "1.0.0",
     "com.unity.xr.magicleap": "4.2.2",
     "com.unity.modules.ai": "1.0.0",
     "com.unity.modules.androidjni": "1.0.0",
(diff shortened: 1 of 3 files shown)


XRTK-Core/Packages/manifest.json
```diff
@@ -15,8 +15,6 @@
     "com.unity.ide.vscode": "1.2.3",
     "com.unity.mobile.android-logcat": "1.2.1",
     "com.unity.test-framework": "1.1.24",
-    "com.unity.textmeshpro": "2.1.4",
-    "com.unity.ugui": "1.0.0",
```
SimonDarksideJ, line 19: Does this not cause issue, removing the Unity UI system?  Granted Core doesn't use it

XRTK-Core/Packages/packages-lock.json
```diff
@@ -62,7 +62,7 @@
     },
     "com.unity.textmeshpro": {
```
SimonDarksideJ, line 63: Are these still needed in the .lock file if they have been removed from the manifest?

StephenHodgson: Lock file is automatically updated by unity and keeps a lost of the direct dependencies in the project.

XRTK-Core/Packages/manifest.json
StephenHodgson: No bc the dependency is defined in the project.json

StephenHodgson: Just means one less place to have to update the information

SimonDarksideJ (approved):

PR #830, opened 2021-05-06
Update Package installer to install boundary config

SimonDarksideJ:

## Overview

Updated the Package installer to also populate Boundary system provider configuration

## Changes

Added extra target for package installer


XRTK-Core/Packages/com.xrtk.core/Editor/PackageInstaller.cs
```diff
@@ -270,6 +272,19 @@ public static void InstallConfiguration(MixedRealityPlatformServiceConfiguration
                             }
                         }
                         break;
+                    case Type _ when typeof(IMixedRealityBoundaryDataProvider).IsAssignableFrom(configurationType):
```
StephenHodgson, line 275: I bet we could make this more generic

Issue #799, opened 2021-03-10
Updated MixedRealityToolkit instance inspector needs to include a "Basic" view, similar to Controllers


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Feature Request

## Is your feature request related to a problem? Please describe

The updated MixedRealityToolkit root profile inspector, whilst improved, does lack a certain amount of usability and could benefit from a "Basic" view, similar to that used in Controler config

## How would you classify your suggestion

- Usability / Configuration

## Describe the solution you'd like
The current screen

![image](https://user-images.githubusercontent.com/1793042/110710573-0625a380-81f6-11eb-985c-b8d257c16e27.png)

Whilst functional for advanced users, could be simplified with a basic view that just has a button to view each systems registered configuration, in the same way with controllers, we have a basic view to just view the mappings for each controller as shown below:

![image](https://user-images.githubusercontent.com/1793042/110710677-353c1500-81f6-11eb-988b-6d946ca4cdc4.png)

## Describe alternatives you've considered
Current screen

SimonDarksideJ:
Current status

Advanced
-----------
![Snag_369e27a](https://user-images.githubusercontent.com/1793042/120403584-62531c00-c33c-11eb-8eea-35e921c09518.png)

Basic
------
![Snag_369f43d](https://user-images.githubusercontent.com/1793042/120403593-64b57600-c33c-11eb-9861-7e14ff57fbf9.png)

Only thing left is the minor task of getting the editor to actually open the profile in the basic view :D

SimonDarksideJ:
I'm no designer, we can update the textures to whatever we want.  I'll see what I can do about the sizing.

StephenHodgson:
I think we should still have a way to assign a profile in at all times

SimonDarksideJ:
If you want to change the current profile, then use the advanced view, simples

StephenHodgson:
I also think the advanced column is a big misleading visually bc people will think it's the profiles off to the side that are advanced. Could we get another screenshot of then taking up a whole row each?

SimonDarksideJ:
Updated look and feel from review comments. 
![Snag_5aa17a](https://user-images.githubusercontent.com/1793042/120468890-a1638a80-c399-11eb-9465-d07f92952d50.png)

SimonDarksideJ:
WHen I tested using the entire Row @StephenHodgson , it looked very clunky and wasteful

StephenHodgson:
[...]

PR #848, opened 2021-06-02
clickable services without having to open foldout

StephenHodgson:

## Overview
Added the ability to click on services to open their configurations with less clicks. should work automatically for all services and data providers.
D3FdIz6vLi

## Changes

- fixes: #799 
- fixed an invalid cast exception when navigating back to the input system profile from a controller data provider profile
- ensured that the foldout stays open when a profile is expected in the service configuration
- fixed an issue in the input system profile where the "Gaze Cursor Prefab" label would disappear when the `IMixedRealityInputDataProvider` property was closed.

## Submodule Changes

- https://github.com/XRTK/Examples/pull/41
- https://github.com/XRTK/Lumin/pull/114
- https://github.com/XRTK/WindowsMixedReality/pull/142


FejZa (approved):
Looks good. I'd like Simon to check as well. Not sure how this happened but looks to me like two people invested time in addressing the same thing and we ended up with two solutions.

SimonDarksideJ (approved):

SimonDarksideJ:
It is frustrating that several paths were developed for the same solution. We just need to get better/more serious about feeding back into feature request tasks to avoid duplication. (and not just dismiss because we aren't interested)

We need to evolve requirements about Requests with any relevant views with Constructive feedback (solutions not dismissals)

StephenHodgson:
Understandable. From a purely objective criticism, I was worried the changes were too broad without adequately addressing all the services and data providers. The changes here are targeted and doesn't require us to add additional classes or icon assets.

PR #868, opened 2021-07-24
Data Provider tests for Core Systems and Extension Services updated

SimonDarksideJ:

## Overview

Updated tests to include validation for data providers registered with their host services, for both "Systems" and "Extension Services"

## Changes

Service Tests

- Test Service 1 - updated with data provider registration logic, matching core systems
- Test DataProvider1 - Updated to register with TestService1 on initialisation
- Test 1 - ServicePlusDataProvider - Updated to validate the count of data providers registered
- Test 1 - ServicePlusDataProvider - Updated to validate the data providers registered to match the registered Test DataProvider

## Notes

Currently failing as there is an issue with Extension Data Providers being registered with their respective service.
Depends on
- https://github.com/XRTK/XRTK-Core/pull/871


XRTK-Core/Packages/com.xrtk.core/Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -617,6 +624,43 @@ public void Test_05_09_GetAllMixedRealityExtensionServices()
...
+        [Test]
+        public void Test_05_10_RegisterMixedRealityExtensionServiceAndDataProvider()
+        {
+            TestUtilities.InitializeMixedRealityToolkitScene(false);
+
+            var activeSystemCount = MixedRealityToolkit.ActiveSystems.Count;
+            var activeServiceCount = MixedRealityToolkit.RegisteredMixedRealityServices.Count;
+
+            // Register
+            MixedRealityToolkit.TryRegisterService<ITestExtensionService1>(new TestExtensionService1("Test17-1.1"));
```
StephenHodgson, line 636: Could we use a more meaningful name for the service?

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionDataProvider1.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace XRTK.Tests.Services
 {
-    internal interface ITestExtensionDataProvider1 : IMixedRealityExtensionDataProvider, ITestService { }
+    internal interface ITestExtensionDataProvider1 : IMixedRealityExtensionDataProvider { }
```
StephenHodgson, line 8: Why did we remove the ITestService interface from all the classes?

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionService1.cs
```diff
@@ -1,9 +1,30 @@
...
+        /// The list of <see cref="IMixedRealityExtensionDataProvider"/>s registered and running with the system.
+        /// </summary>
+        IReadOnlyCollection<IMixedRealityExtensionDataProvider> DataProviders { get; }
+
+        /// <summary>
+        /// Registers the <see cref="ITestExtensionDataProvider1"/> with the <see cref="ITestService"/>.
+        /// </summary>
+        /// <param name="dataProvider"></param>
+        /// <returns></returns>
+        bool RegisterDataProvider(ITestExtensionDataProvider1 dataProvider);
```
StephenHodgson, line 21: Seems kinda off that this is here and not being properly registered via the xrtk service registry

XRTK-Core/Packages/com.xrtk.core/Tests/Services/TestDataProvider1.cs
```diff
@@ -11,8 +11,11 @@ internal class TestDataProvider1 : BaseDataProvider, ITestDataProvider1
         public TestDataProvider1(ITestService parentService, string name = "Test Data Provider 1", uint priority = 1, BaseMixedRealityProfile profile = null)
             : base(name, priority, profile, parentService)
         {
+            testSystem = parentService;
```
StephenHodgson, line 14: Where is test system defined?

XRTK-Core/Packages/com.xrtk.core/Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
SimonDarksideJ: It is using the naming conventions set down by the initial person who wrote the tests.  As it is just a test, the naming is not that critical, other than to state it is an "Extension" service.

If you wish to change all the names of all the definitions, please feel free to submit a PR

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionService1.cs
SimonDarksideJ: This is the implementation used by ALL of the current services in the XRTK, so the test was updated to reflect how it is currently implemented.

Please feel free to submit a PR and change the core systems and then the tests if you don't like it.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/TestDataProvider1.cs
SimonDarksideJ: On Line 17.
The pattern was replicated from the existing systems active in use in the XRTK.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionDataProvider1.cs
SimonDarksideJ: Because ITestService directly inherits from IMixedRealityService
```
ITestService : IMixedRealityService
```

This is an extension service, so should inherit from 'IMixedRealityExtensionService', else it is not a valid test of an ExtensionService.

SimonDarksideJ: Even crazier, was the fact the data providers were configured to inherit directly from IMixedRealityService, which is not a real-world test case.

SimonDarksideJ:
If there is an issue with the test, please detail it, as the Extension Service and Extension Data Provider have been updated to match what I understand to be the correct pattern, inheriting from the Extension Interfaces.
The data provider use has also been updated in the Core and Extension tests to match the real-world implementation we currently have running in our systems (although granted, there is scope for improvement there).

The Core System implementation and it's data providers pass their tests.
The same implementation using the Extension interfaces does NOT pass.

Either we need to fix the tests, or change the XRTK extension implementation as it is not currently working as expected.

XRTK-Core/Packages/com.xrtk.core/Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
StephenHodgson: I just think 17-1.1 doesn't really make sense. The test isn't even the 17th test. Just curious where 17-1.1 came from is all and how it relates to the name of the test service instance.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionDataProvider1.cs
StephenHodgson: Ah I see, it should probably be inheriting from a `ITestDataProvider` instead of a service. great catch!

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionService1.cs
StephenHodgson: Yeah but I don't think it's actually part of any interface, just a pattern we tend to use, which is why I asked. I think this might benefit from either adding to the service interface and/or base service class.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/TestDataProvider1.cs
StephenHodgson: Oh idk I missed that 😅

XRTK-Core/Packages/com.xrtk.core/Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
StephenHodgson: imho since it's test 5.10 it almost makes more sense if the name was `Test5.10.1` or something

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionService1.cs
```diff
@@ -1,9 +1,30 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Collections.Generic;
 using XRTK.Interfaces;
 
 namespace XRTK.Tests.Services
 {
-    internal interface ITestExtensionService1 : IMixedRealityExtensionService, ITestService { }
+    internal interface ITestExtensionService1 : IMixedRealityExtensionService
```
StephenHodgson, line 9: I think that since this is still an extension service we can keep the previous inheritance from `ITestService`.

We probably want to run the same set of tests against each.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/TestService1.cs
```diff
@@ -24,5 +33,26 @@ public override void Disable()
         {
             IsEnabled = false;
         }
+
+        public bool RegisterDataProvider(ITestDataProvider1 dataProvider)
```
StephenHodgson, line 37: I'm wondering if there's a better way to do this.

I don't really like the pattern we're using even in the real services and data providers. Any data provider already registered under the service should technically already be known about. As long as it matches one of the pre-registered types why does it need to register itself again, ya know?

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestService.cs
```diff
@@ -1,12 +1,33 @@
...
 namespace XRTK.Tests.Services
 {
     internal interface ITestService : IMixedRealityService
     {
         bool IsEnabled { get; }
+
+        /// <summary>
+        /// The list of <see cref="IMixedRealityDataProvider"/>s registered and running with the system.
+        /// </summary>
+        IReadOnlyCollection<IMixedRealityDataProvider> DataProviders { get; }
```
StephenHodgson, line 17: @FejZa @SimonDarksideJ Thoughts about promoting this to `IMixedRealityService`?

XRTK-Core/Packages/com.xrtk.core/Tests/Services/TestService1.cs
SimonDarksideJ: That was my thinking, I hate boilerplate, as this registration pattern (which is valid, data providers should announce themselves to their service), but is there a better way so we don't need to declare it everytime

XRTK-Core/Packages/com.xrtk.core/Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
SimonDarksideJ: The numbering simply followed on from the numbering you implemented in the previous test.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionDataProvider1.cs
SimonDarksideJ: Except it shouldn't be inheriting from something that isn't for "Extensions", hence 

ExtensionService -> ExtensionDataProvider.

Shouldn't have anything to do with the base versions of Data Provider, unless we are dropping the "Extension" segment for Data Providers, which wouldn't be the worst thing.  A data provider is just a data provider, it's the Service/System that is the extension.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionService1.cs
SimonDarksideJ: Agreed there, I hate boilerplate if it can be genericized in the base somewhere for dataproviders to call.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionDataProvider1.cs
StephenHodgson: I mean there's really no difference between the extensions vs non extensions at the moment (Maybe in the future?), Which is why I'm confused about how and why we might be seeing different results.

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestService.cs
SimonDarksideJ: In a separate PR, I am going to unweave the RegisteredServiceProviders

XRTK-Core/Packages/com.xrtk.core/Tests/Services/TestService1.cs
SimonDarksideJ: This needs to be addressed in a separate PR

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionService1.cs
SimonDarksideJ: Extension Services are going

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionService1.cs
SimonDarksideJ: Needs to be addressed in a separate PR

XRTK-Core/Packages/com.xrtk.core/Tests/Services/ITestExtensionDataProvider1.cs
SimonDarksideJ: Extension Services to be removed in a separate PR

SimonDarksideJ:
Any ideas what this is?

```
Packages\com.xrtk.core\Tests\Services\TestService1.cs(5,19): error CS0234: The type or namespace name 'VersionControl' does not exist in the namespace 'UnityEditor' (are you missing an assembly reference?)
```

We haven't changed the packages.  I'll look at it again in a short while

XRTK-Core/Packages/com.xrtk.core/Tests/Services/TestDataProvider1.cs
```diff
@@ -2,14 +2,20 @@
...
 namespace XRTK.Tests.Services
 {
     internal class TestDataProvider1 : BaseDataProvider, ITestDataProvider1
     {
+        /// <inheritdoc />
         public TestDataProvider1(ITestService parentService, string name = "Test Data Provider 1", uint priority = 1, BaseMixedRealityProfile profile = null)
             : base(name, priority, profile, parentService)
+
+        //public TestDataProvider1(string name, uint priority, BaseMixedRealityProfile profile, ITestService parentService)
+        //    : base(name, priority, profile, parentService)
```
StephenHodgson, line 16: why commented out?

SimonDarksideJ:
Will rebuild without dependencies on old PR's

PR #871, opened 2021-08-01
Remove extension services

SimonDarksideJ:

## Overview

Removed extension services

## Changes

- Removed Extension Definitions / Interfaces / Implementatons
- Renamed "RegisteredServiceProviders" to "ActiveSystemsDataProviders"
- Removed "RegisteredServiceProviders" Profiles and definitions against the "MixedRealityToolkit" service
- Removed Extensions Tests
- General cleanup and tidy
- Updated https://github.com/XRTK/Examples/pull/45

## Breaking Changes

- Breaks Any implementation using 'Extension' interfaces, should be converted to Systems or Services


XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
```diff
@@ -162,15 +162,15 @@ public void ResetProfile(MixedRealityToolkitRootProfile profile)
         public static IReadOnlyDictionary<Type, IMixedRealitySystem> ActiveSystems => activeSystems;
 
         // ReSharper disable once InconsistentNaming
-        private static readonly List<Tuple<Type, IMixedRealityService>> registeredMixedRealityServices = new List<Tuple<Type, IMixedRealityService>>();
+        private static readonly List<Tuple<Type, IMixedRealityService>> activeSystemsDataProviders = new List<Tuple<Type, IMixedRealityService>>();
```
StephenHodgson, line 165: I think we should keep the name the same

SimonDarksideJ: But they are not RegisteredServices and will never contain such?

XRTK-Core/Packages/com.xrtk.core/Editor/Templates~/SystemProfile.txt
```diff
@@ -2,7 +2,11 @@
 
 namespace #NAMESPACE#
 {
+#if UNITY_EDITOR
```
FejZa, line 5: No need to put this in #if UNITY_EDITOR

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
FejZa: I like both names really. This is an internal private field anyways, so it's not as important as an public API. Even if it was, both names work for me. So I am not sure why I'm even making a comment :)

StephenHodgson: Data providers is too specific. Should continue to be services

StephenHodgson (changes requested):
Please revert name change for registered mixed reality services

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
SimonDarksideJ: Ok, fair enough, I'll rename it back

SimonDarksideJ: Reverted

XRTK-Core/Packages/com.xrtk.core/Editor/Templates~/SystemProfile.txt
SimonDarksideJ: Removed

XRTK-Core/Packages/com.xrtk.core/Editor/Templates~/SystemProfile.txt
```diff
@@ -2,7 +2,9 @@
 
 namespace #NAMESPACE#
 {
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/#NAME#/#NAME#Profile", fileName = "#NAME#Profile", order = 99)]
```
StephenHodgson, line 5: I intentionally left this out before because I didn't think people would want their profiles showing up under the toolkit menu

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/MixedRealityRegisteredServiceProvidersProfile.cs
```diff
@@ -1,14 +0,0 @@
...
-// Licensed under the MIT License. See LICENSE in the project root for license information.﻿
-
-using UnityEngine;
-using XRTK.Definitions.Utilities;
-using XRTK.Interfaces;
-
-namespace XRTK.Definitions
-{
-    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Registered Service Providers Profile", fileName = "MixedRealityRegisteredServiceProvidersProfile", order = (int)CreateProfileMenuItemIndices.RegisteredServiceProviders)]
-    public class MixedRealityRegisteredServiceProvidersProfile : BaseMixedRealityServiceProfile<IMixedRealityExtensionService>
```
StephenHodgson, line 11: I think we should keep this class but change the interface to `IMixedRealityService`

XRTK-Core/Packages/com.xrtk.core/Tests/InputSystem/TestFixture_03_InputSystemTests.cs
```diff
@@ -16,7 +16,6 @@ public void Test01_CreateMixedRealityInputSystem()
         {
             TestUtilities.InitializeMixedRealityToolkitScene(false);
             var activeSystemCount = MixedRealityToolkit.ActiveSystems.Count;
-            var activeServiceCount = MixedRealityToolkit.RegisteredMixedRealityServices.Count;
```
StephenHodgson, line 19: let's revert this check

XRTK-Core/Packages/com.xrtk.core/Tests/InputSystem/TestFixture_03_InputSystemTests.cs
```diff
@@ -26,7 +25,6 @@ public void Test01_CreateMixedRealityInputSystem()
             Assert.IsNotEmpty(MixedRealityToolkit.ActiveSystems);
             Assert.IsNotNull(inputSystem);
             Assert.AreEqual(activeSystemCount + 1, MixedRealityToolkit.ActiveSystems.Count);
-            Assert.AreEqual(activeServiceCount + 1, MixedRealityToolkit.RegisteredMixedRealityServices.Count);
```
StephenHodgson, line 29: let's revert this check

XRTK-Core/Packages/com.xrtk.core/Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -705,11 +461,9 @@ public void Test_07_01_EnableServicesByType()
 
             // Add test 1 services
             MixedRealityToolkit.TryRegisterService<ITestDataProvider1>(new TestDataProvider1(testService1));
-            MixedRealityToolkit.TryRegisterService<ITestExtensionService1>(new TestExtensionService1());
```
StephenHodgson, line 708: shouldn't this just be:

```
MixedRealityToolkit.TryRegisterService<ITestService1>(new TestService1());
```

Seems like odd that we're just removing instead of updating the usages. I'd assume it should work just the same since there wasn't really any functional difference between an extension service and a regular one.

StephenHodgson (changes requested):
See comments

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
StephenHodgson: Thanks. Remember that System implies a single instance, where as a service implies there could be many.

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/MixedRealityRegisteredServiceProvidersProfile.cs
FejZa: Can you clarify? We can register services without this profile in the root profile? I thought it's the whole point.

XRTK-Core/Packages/com.xrtk.core/Editor/Templates~/SystemProfile.txt
SimonDarksideJ: Except without it, the Profile is pointless as you can't create it.
We can always have a guide that tells you to rename it.

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/MixedRealityRegisteredServiceProvidersProfile.cs
SimonDarksideJ: I see your point about having a "default" inspector @StephenHodgson , however, with nothing configured nothing will draw anyway.  So what is the point (other than it will draw the XRTK logo.)

XRTK-Core/Packages/com.xrtk.core/Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
SimonDarksideJ: Renamed, although I could <blame> the original test creator for that ;D

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/MixedRealityRegisteredServiceProvidersProfile.cs
StephenHodgson: The root configuration profile only shows registered systems, which implies a single instance. We don't have a dedicated place to register ad hoc services.

SimonDarksideJ: I'll update the Path in the CreateAssetMenu to be more specific 
But ultimately:

* Custom Systems - Are registerable through the editor along with their profiles
* Custom Services - Are not, and are only accessible through code.

Both have their uses.

XRTK-Core/Packages/com.xrtk.core/Editor/Templates~/SystemProfile.txt
SimonDarksideJ: I'll change it so that the path is not XRTK and is specific to the service being generated.   The Developer can then chose to update/change it.

SimonDarksideJ: Switched to 

> [CreateAssetMenu(menuName = "#NAMESPACE#/#NAME#/#NAME#Profile", fileName = "#NAME#Profile", order = 99)]

XRTK-Core/Packages/com.xrtk.core/Runtime/Definitions/MixedRealityRegisteredServiceProvidersProfile.cs
SimonDarksideJ: Restored and updated

SimonDarksideJ:
Any final comments @StephenHodgson  / @FejZa would love to get this cleared up asap.

StephenHodgson:
I'll give it a once over today

XRTK-Core/Packages/com.xrtk.core/Editor/Profiles/MixedRealityToolkitRootProfileInspector.cs
```diff
@@ -21,9 +21,6 @@ namespace XRTK.Editor.Profiles
     [CustomEditor(typeof(MixedRealityToolkitRootProfile))]
     public class MixedRealityToolkitRootProfileInspector : MixedRealityServiceProfileInspector
     {
-        // Additional registered components profile
-        private SerializedProperty registeredServiceProvidersProfile;
```
StephenHodgson, line 25: I would like to put this back in please. That way there's a UI to register services that aren't systems

StephenHodgson (changes requested):
Please revert removal of registered service profile

XRTK-Core/Packages/com.xrtk.core/Editor/Profiles/MixedRealityToolkitRootProfileInspector.cs
SimonDarksideJ: Now this I don't get.

We remove Extension Services, so there is a single unified collection of Services/Systems through the inspector.

But now you want to retain the separate profile we were removing so "other" services can be configured.

Surely we should have a better plan where ALL services are in a single place and not hidden behind other config

FejZa: @SimonDarksideJ I think I understand what Stephen is going for. Say you remove that profile, then in order to to register any custom services, I'll have to first create a "dummy" system and register it in the root profile. Only then I will be able to add/register services to that system via the inspector.

So basically this `RegisteredServiceProvidersProfile` is that "dummy" system. It allows me to register services without having to worry abotu setting up a system first. Does that make sense to you?

FejZa: Remember system = only one active implementation/instance. Service = multiple instances allowed.

FejZa: I think you did a great job removing the extension services defintions. As we don't need them. You united all services to be the same, which is a win. But I do also believe we can keep this profile just so people have a way to register services without having to worry about setting up a system first.

SimonDarksideJ: Which is fair enough, but as you can have a system that can have MANY data providers, isn't that a better approach than multiple services?  Using more of a "Pool" approach.

StephenHodgson: Remember this diagram?
0_MTUg71o438oMaOBv

SimonDarksideJ:
Updated as requested.

However, note.  The CUstom Service example will need to be updated accordingly, as it does not currently work.  The editor reports a "Constuctor Error" when trying to instantiate a "Custom Service" that has no profile or constructor.

StephenHodgson:
> However, note. The Custom Service example will need to be updated accordingly, as it does not currently work. The editor reports a "Constructor Error" when trying to instantiate a "Custom Service" that has no profile or constructor.

I'll take a look when I get a chance.

SimonDarksideJ:
Any last comments @StephenHodgson  / @FejZa ?

StephenHodgson:
Sorry haven't had time to check this out

XRTK-Core/Packages/com.xrtk.core/Editor/Templates~/SystemProfile.txt
```diff
@@ -2,7 +2,9 @@
 
 namespace #NAMESPACE#
 {
+    [CreateAssetMenu(menuName = "#NAMESPACE#/#NAME#/#NAME#Profile", fileName = "#NAME#Profile", order = 99)]
     public class #NAME#Profile : #BASE#
     {
     }
+
```
StephenHodgson, line 9: nit extra line

SimonDarksideJ: You do realise you could have just "done" it.

StephenHodgson:
Do you guys think we could push this off into 0.4.x release?

FejZa:
I am fine with that. Let's let @SimonDarksideJ have the last word on this one.

PR #875, opened 2021-08-15
Remove internal flag on parent profile property

FejZa:

## Overview

Removed the `internal` keyword on the `BaseMixedRealityProfile.ParentProfile` property setter. This was necessary to be able to set parent profiles  in code that lives in another assembly.


StephenHodgson (changes requested):
This was internal on purpose. Can you better describe the use case?

FejZa:
What was the purpose of making it internal? Please see https://github.com/XRTK/Examples/pull/46 for use case.

StephenHodgson:
Please use  use `InternalsVisibleTo()` assembly attribute instead of making the setter public

FejZa:
Done

StephenHodgson (approved):

PR #876, opened 2021-08-16
New Locomotion System

FejZa:

## Overview

This PR introduces the new locomotion system which replaces the old teleport system. This new system is set up in a way that it is extremely easy for users to implement new types of locomotion if needed but also comes with common default implementations for smooth locomotion and teleportation.

Key Features:
- Change locomotion types at runtime
- Register locomotion implementations as data providers (no MonoBehaviours)
- Very flexible in use and extend

https://user-images.githubusercontent.com/9565734/129541455-ae127fe7-d0a1-4590-b6ce-bbf1f6910485.mp4

## Changes
- Remove Teleport System
- Add Locomotion System

## Breaking Changes
- Teleport System is gone, require updating existing root profiles to register the new Locomotion System

## Related Submodule Changes

- https://github.com/XRTK/SDK/pull/243
- https://github.com/XRTK/Examples/pull/47


StephenHodgson:
Please double check that all the submodules are pointing to the right commits

SimonDarksideJ:
Curious, the UltraLeap changes in that Branch?  should that be in this PR?

FejZa:
Nope. I gotta do some clean up here, not sure what happened. I'll let you know.

SimonDarksideJ:
Bit curious as to the alignment of implementations in base and what is in the SDK.

I was under the impression we were just going to have a single (default) Locomotion provider in the core and all the others would be in the SDK, am I incorrect?

If so, all inspectors and relevant providers in the SDK should also have their inspectors / profile definitions and so on in the SDK.

Also find it curious the "BaseTeleportProvider" is in the SDK and not core?

FejZa:
We talked about this a while ago. I don't like having to move anything to SDK. All the default implementations are in Core. Especially base classes like `BaseLocomotionProvider` etc. Because if they weren't, then I couldn't make use of them in Core implementations without creating a dependency on SDK. Which does not make sense.

In general, back then when we talked about it, we agreed to move things to Core so they can work on their own without the SDK.

The only thing living in SDK are the default profiles basically.

FejZa:
Oh yeah sorry, Pointers and HotSpot are in SDK as well (which I would love to move to Core tbh) :D

StephenHodgson:
> Pointers and HotSpot

Makes sense to me. 

Yeah the general consensus was to try to keep monobehaviours out of core as much as possible for future engine porting. But at this point I don't think it really matters. I think we should just stick to Unity.

FejZa:
I see. Well I think we can leave it like this for now. And yes I think we should just stick to Unity.

I am really torn in between making locomotion its own package and not.

StephenHodgson:
> I am really torn in between making locomotion its own package and not.

I think it is ubiquitous enough to stay in core

FejZa:
Renamed hotspots to TeleportAnchor

StephenHodgson:
Needs to be fixed. It won't pass until the packages are updated I think

FejZa:
Any idea how I can reset submodules? This PR shows I made changes in Ultraleap and glTF, which does not make any sense. Not sure what happened here.

StephenHodgson:
I can take a quick look

XRTK-Core/Assets/Assets.index
```diff
@@ -0,0 +1,19 @@
+{
```
StephenHodgson, line 1: What does this file do?

XRTK-Core/Packages/com.xrtk.core/Runtime/EventDatum/Diagnostics/BaseDiagnosticsEventData.cs
```diff
@@ -5,7 +5,7 @@
 using UnityEngine.EventSystems;
 using XRTK.Extensions;
 
-namespace XRTK.EventDatum.DiagnosticsSystem
+namespace XRTK.EventDatum.Diagnostics
```
StephenHodgson, line 8: Not sure if we should include this kind of breaking change in with the Locomotion changes.

XRTK-Core/Packages/com.xrtk.core/Runtime/Interfaces/InputSystem/IMixedRealityPointer.cs
```diff
@@ -48,11 +47,6 @@ public interface IMixedRealityPointer : IEqualityComparer
         /// </summary>
         ICursorModifier CursorModifier { get; set; }
 
-        /// <summary>
-        /// The currently active teleport hotspot.
-        /// </summary>
-        IMixedRealityTeleportHotSpot TeleportHotSpot { get; set; }
```
StephenHodgson, line 54: Not replaced with ITeleportAnchor?

FejZa: Huh? Let me check this I updated ALL of the references and specifically searched for the string "Hotspot" and updated.

XRTK-Core/Assets/Assets.index
FejZa: This file is created by the `Quick Search` upm package. I accidentally checked it in.

StephenHodgson (changes requested):
I think the namespace changes are a bit out of scope for this. 

We can address those in the next release cycle

FejZa:
Thanks. I'll revert some of the changes now.

XRTK-Core/Packages/com.xrtk.core/Runtime/Interfaces/InputSystem/IMixedRealityPointer.cs
FejZa: Ah, I misread this. This is fine. Pointers do no longer have a reference to hotspots/anchors in their interface. So this change is correct.

FejZa:
Feedback addressed @StephenHodgson.

XRTK-Core/Packages/com.xrtk.core/Runtime/Interfaces/InputSystem/IMixedRealityPointer.cs
StephenHodgson: Interesting, how do we know which anchor is active for the specific pointer then?

FejZa: The active anchor is available via `ITeleportTargetProvider.cs`, which happens to be implemented by the `TeleportPointer`. Generally providing a teleport target is not tied to pointers. Anything implementing the mentioned interface can provide teleport targets.

StephenHodgson: This was a convenience property for those consuming pointer event data, so they knew which specfic anchor the pointer was targeting

StephenHodgson: > The active anchor is available via ITeleportTargetProvider.cs

Sounds an awful lot like that pointer mediator 😂

I'll late a look and get back to you.

FejZa: It does not. It's simply an interface for components providing a teleport target. You could be throwing a ball and where it lands that would be your teleport target. See how that is not a pointer?

FejZa: You need to let go of the idea that only pointers provide teleportation targets.

StephenHodgson: @FejZa That's a fair point 😅

StephenHodgson:
LGTM, hopefully the builds start passing again then we'll merge it in

StephenHodgson:
@FejZa I think I figured it out. I reverted the gltf submodule changes. It seems to hang the editor when importing models for some reason. I'll look into it.

StephenHodgson (approved):

FejZa:
Great, thanks. Once the package is published I need to bump SDK dependencies so the build can pass.

StephenHodgson:
It's already published

PR #894, opened 2021-09-30
Enable scriptable render pipeline support for Blink Teleport Provider

FejZa:

## Overview

I need the blink teleport provider to support scriptable render pipelines. In the legacy pipeline it was possible to change a materials properties to support transparency at runtime. With render pipelines this is not possible anymore for performance reasons (they said). So I made the material used for blinking a configurable setting in the provider's profile, which is useful in many ways anyways. Also exposed two more settings that were hardwired in the provider previously.

image

## Changes

- Exposed fade material and colors as settings in the blink teleport provider profile

## Related Submodule Changes

- https://github.com/XRTK/com.xrtk.sdk/pull/246


StephenHodgson (approved):

StephenHodgson:
@FejZa don't forget to update your sdk submodule commit

FejZa:
Thanks, just did

StephenHodgson (approved):

Issue #916, opened 2022-01-31
Console is being spammed with "Infinite Reload" error messages on the Development branch


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug

Even on a fresh clone, opening the XRTK development branch in any version of Unity, the console is being spammed with "Infinite Import loop" error messages.

DOES NOT OCCUR in a project that consumes the built UPM packages.

## To Reproduce

1. Create a fresh clone of the XRTK, recursive.
2. Open Unity in the XRTK project
3. Look at console

## Expected behavior

No Infinite import loop error

## Actual behavior

See error below

## Screenshots

```
An infinite import loop has been detected. The following Assets were imported multiple times, but no changes to them have been detected. Please check if any custom code is trying to import them:
Packages/com.unity.xr.oculus/Third Party Notices.md
Packages/com.needle.compilation-visualizer/package.json
Packages/com.unity.xr.arsubsystems/CHANGELOG.md
Packages/com.unity.xr.magicleap/Third Party Notices.md
Packages/com.xrtk.sdk/Runtime/StandardAssets/Materials/Occlusion.mat
Packages/com.xrtk.ultraleap/CHANGELOG.md
Packages/com.unity.modules.tilemap/package.json
Assets/XR/Loaders/Oculus Loader.asset
Packages/com.unity.xr.magicleap/Tests/Editor/Manifest/Privileges_0.20.0.txt
Packages/com.xrtk.sdk/Runtime/StandardAssets/Materials/Standard/Standard_TransparentPink.mat
Assets/XRTK.Generated/Lumin/Profiles
Packages/com.xrtk.sdk/Runtime/StandardAssets/Materials/Standard/Standard_Magenta.mat
Packages/com.unity.xr.legacyinputhelpers/CHANGELOG.md
Assets/XRTK.Generated/BuildInfo/BuildInfo.asset
Packages/com.unity.xr.management/LICENSE.md
Packages/com.unity.mobile.android-logcat/LICENSE.md
Packages/com.xrtk.core/Runtime/StandardAssets/Materials/3DTextSelawikLight.mat
Packages/com.xrtk.core/Editor/Data/EditorWindowOptions.json
Packages/com.unity.ugui/package.json
Packages/com.xrtk.wmr/package.json
Packages/com.unity.xr.management/Tests/TestTooling/README.md
Packages/com.unity.scriptablebuildpipeline/README.md
[...]

PR #917, opened 2022-01-31
Remove cached symlink files which can cause errors on fresh clones

SimonDarksideJ:

## Overview

In some circumstances, the cached symlink files can cause issues on machines by becoming read-only, this causes issues such as identified in #916 
As the Symlinks are intended to be autogenerated links from other sources, these cached files are not needed.

The GitIgnore file has been updated to ensure these links are not added again.

Any required changes to the XRTK.Generated folder should be manually added in the future.

## Changes

- Fixes: [](https://github.com/XRTK/com.xrtk.core/issues/916)

## Breaking Changes

None


StephenHodgson (changes requested):
This will break things. 

This was done by design. The issue is with the computer env setup with Git Settings. Check setup documentation

Issue #911, opened 2022-01-18
Console being spammed with XRRig warnings


SimonDarksideJ:
# XRTK - Mixed Reality Toolkit Bug Report

## Describe the bug
In both a clean project and in XRTK dev, with a scene using a configured XRCameraRig, the toolkit seems to be constantly trying to recreate the rig and in some cases creates multiple.

This causes the error described below to be spammed in to the console.

## To Reproduce

(Only tested on Oculus / Android / Standalone atm)

1. Clone Development
2. Open Unity for XRTK-Core project
3. Create a new Scene
4. Use -> MixedRealityToolkit -> configure
5. View MRTK config, especially camera profile
6. Observe Console.

## Expected behavior

No warnings if the scene has not changed

## Actual behavior

Spam

## Screenshots
```
The Mixed Reality Toolkit expected the camera's parent to be named XRCameraRig. The existing parent will be renamed and used instead.
UnityEngine.Debug:LogWarning (object)
XRTK.Services.CameraSystem.DefaultCameraRig:get_PlayerCamera () (at Packages/com.xrtk.core/Runtime/Services/CameraSystem/DefaultCameraRig.cs:125)
XRTK.Services.CameraSystem.DefaultCameraRig:get_CameraTransform () (at Packages/com.xrtk.core/Runtime/Services/CameraSystem/DefaultCameraRig.cs:85)
XRTK.Providers.CameraSystem.BaseCameraDataProvider:UpdateCameraHeightOffset (single) (at Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs:430)
XRTK.Providers.CameraSystem.BaseCameraDataProvider:SetupTrackingOrigin () (at Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs:333)
XRTK.Providers.CameraSystem.BaseCameraDataProvider:Initialize () (at Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs:169)
XRTK.Services.MixedRealityToolkit:InitializeAllServices () (at Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs:1313)
UnityEditor.EditorApplication:Internal_CallDelayFunctions ()

```

Error 2 caused during build & run

```
Some objects were not cleaned up when closing the scene. (Did you spawn new GameObjects from OnDestroy?)
The following scene GameObjects were found:
[...]

PR #919, opened 2022-02-05
Updated Camera validation

SimonDarksideJ:

## Overview

Updated the camera validation and regeneration as a result of #911 

## Changes

* Moved Camera validation routines to "OnEnable" in the Camera Provider
* Moved Rig Validation to the OnValidate method, so it is checked at the right time
* Added Rig Validation to the Toolkit Validation methods
* Ensured Toolkit validation ONLY occurs when a profile has been assigned
* Fixed a random bug whereby CameraCache was being called in OnDestroy without first checking there was a camera to destroy (as it created a new camera)
* Updated validation warnings

Tested all scenarios where the scene will be validated:

* New Scene
* New Scene + Configure
* New GO + Toolkit object
* Deleting Toolkit and readding
* Deleting Rig and checking configuration

Fixes: 
- https://github.com/XRTK/com.xrtk.core/issues/911

## Breaking Changes

None

### DIFF
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
@@ -308,7 +308,7 @@ namespace XRTK.Providers.CameraSystem
             cameraSystem.UnRegisterCameraDataProvider(this);
         }
 
-#endregion IMixedRealitySerivce Implementation
+        #endregion IMixedRealitySerivce Implementation
 
         private void EnsureCameraRigSetup()
         {
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
@@ -168,8 +168,8 @@ namespace XRTK.Providers.CameraSystem
 #if XRTK_USE_LEGACYVR
             ApplySettingsForDefaultHeadHeight();
 #else
-            // We attempt to intialize the camera tracking origin, which might
-            // fail at this point if the subystems are not ready, in which case,
+            // We attempt to initialize the camera tracking origin, which might
+            // fail at this point if the subsytems are not ready, in which case,
             // we set a flag to keep trying.
             trackingOriginInitialized = SetupTrackingOrigin();
             trackingOriginInitializing = !trackingOriginInitialized;
@@ -245,9 +245,9 @@ namespace XRTK.Providers.CameraSystem
             }
 
 #if !XRTK_USE_LEGACYVR
-            // We keep trying to intiailze the tracking origin,
+            // We keep trying to initialize the tracking origin,
             // until it worked, because at application launch the
-            // subystems might not be ready yet.
+            // subsytems might not be ready yet.
             if (trackingOriginInitializing && !trackingOriginInitialized)
             {
                 trackingOriginInitialized = SetupTrackingOrigin();
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
@@ -158,8 +158,41 @@ namespace XRTK.Providers.CameraSystem
 
             cameraSystem.RegisterCameraDataProvider(this);
 
+            if (!Application.isPlaying)
+            {
+                return;
+            }
 
+            EnsureCameraRigSetup();
 
+#if XRTK_USE_LEGACYVR
+            ApplySettingsForDefaultHeadHeight();
+#else
+            // We attempt to intialize the camera tracking origin, which might
+            // fail at this point if the subystems are not ready, in which case,
+            // we set a flag to keep trying.
+            trackingOriginInitialized = SetupTrackingOrigin();
+            trackingOriginInitializing = !trackingOriginInitialized;
+#endif
+
+            cameraOpaqueLastFrame = IsOpaque;
+
+            if (applyQualitySettings)
+            {
+                if (IsOpaque)
+                {
+                    ApplySettingsForOpaqueDisplay();
+                }
+                else
+                {
+                    ApplySettingsForTransparentDisplay();
+                }
+            }
+
+            if (Application.isPlaying)
+            {
+                XRSettings.eyeTextureResolutionScale = eyeTextureResolution;
+            }
         }
 
         /// <inheritdoc />
@@ -290,36 +323,6 @@ namespace XRTK.Providers.CameraSystem
                 CameraRig = CameraCache.Main.transform.parent.gameObject.EnsureComponent(cameraRigType) as IMixedRealityCameraRig;
                 Debug.Assert(CameraRig != null);
             }
-
-#if XRTK_USE_LEGACYVR
-            ApplySettingsForDefaultHeadHeight();
-#else
-            // We attempt to intialize the camera tracking origin, which might
-            // fail at this point if the subystems are not ready, in which case,
-            // we set a flag to keep trying.
-            trackingOriginInitialized = SetupTrackingOrigin();
-            trackingOriginInitializing = !trackingOriginInitialized;
-#endif
-
-            cameraOpaqueLastFrame = IsOpaque;
-
-            if (applyQualitySettings)
-            {
-                if (IsOpaque)
-                {
-                    ApplySettingsForOpaqueDisplay();
-                }
-                else
-                {
-                    ApplySettingsForTransparentDisplay();
-                }
-            }
-
-
-            if (Application.isPlaying)
-            {
-                XRSettings.eyeTextureResolutionScale = eyeTextureResolution;
-            }
         }
 
 #if !XRTK_USE_LEGACYVR
@@ -334,7 +337,6 @@ namespace XRTK.Providers.CameraSystem
             // we can still keep going and assume everything is ready.
             var trackingOriginModeSet = true;
 
-#if !UNITY_EDITOR
             if (inputSubsystems.Count != 0)
             {
                 for (int i = 0; i < inputSubsystems.Count; i++)
@@ -355,10 +357,6 @@ namespace XRTK.Providers.CameraSystem
                 // connected, position the camera at the configured default offset.
                 UpdateCameraHeightOffset(defaultHeadHeight);
             }
-#else
-            // The editor does not have a tracking type, so just update the default headheight for simulation
-            UpdateCameraHeightOffset(defaultHeadHeight);
-#endif 
 
             return trackingOriginModeSet;
         }
@@ -449,9 +447,9 @@ namespace XRTK.Providers.CameraSystem
 #endif
 
 #if !XRTK_USE_LEGACYVR
-            /// <summary>
-            /// Updates the camera height offset to the specified value.
-            /// </summary>
+        /// <summary>
+        /// Updates the camera height offset to the specified value.
+        /// </summary>
         protected virtual void UpdateCameraHeightOffset(float heightOffset = 0f)
         {
             CameraRig.CameraTransform.localPosition = new Vector3(
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
@@ -341,14 +341,20 @@ namespace XRTK.Providers.CameraSystem
             {
                 for (int i = 0; i < inputSubsystems.Count; i++)
                 {
-                    var result = SetupTrackingOrigin(inputSubsystems[i]);
-                    if (result)
+                    if (inputSubsystems[i].SubsystemDescriptor.id == "MockHMD Head Tracking")
                     {
-                        inputSubsystems[i].trackingOriginUpdated -= XRInputSubsystem_OnTrackingOriginUpdated;
-                        inputSubsystems[i].trackingOriginUpdated += XRInputSubsystem_OnTrackingOriginUpdated;
+                        UpdateCameraHeightOffset(defaultHeadHeight);
                     }
-
-                    trackingOriginModeSet &= result;
+                    else
+                    {
+                        var result = SetupTrackingOrigin(inputSubsystems[i]);
+                        if (result)
+                        {
+                            inputSubsystems[i].trackingOriginUpdated -= XRInputSubsystem_OnTrackingOriginUpdated;
+                            inputSubsystems[i].trackingOriginUpdated += XRInputSubsystem_OnTrackingOriginUpdated;
+                        }
+                        trackingOriginModeSet &= result;
+                    }
                 }
             }
             else
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
@@ -316,7 +316,7 @@ namespace XRTK.Providers.CameraSystem
             {
                 if (CameraCache.Main.transform.parent.IsNull())
                 {
-                    var rigTransform = new GameObject().transform;
+                    var rigTransform = new GameObject(MixedRealityToolkit.DefaultXRCameraRigName).transform;
                     CameraCache.Main.transform.SetParent(rigTransform);
                 }
 
@@ -420,7 +420,7 @@ namespace XRTK.Providers.CameraSystem
             SyncRigTransforms();
         }
 
-#endregion Tracking Origin Setup
+        #endregion Tracking Origin Setup
 #endif
 
 #if XRTK_USE_LEGACYVR
--- a/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
+++ b/XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
@@ -156,54 +156,18 @@ namespace XRTK.Providers.CameraSystem
         {
             base.Initialize();
 
-            if (CameraRig == null)
-            {
-                if (CameraCache.Main.transform.parent.IsNull())
-                {
-                    var rigTransform = new GameObject().transform;
-                    CameraCache.Main.transform.SetParent(rigTransform);
-                }
-
-                CameraRig = CameraCache.Main.transform.parent.gameObject.EnsureComponent(cameraRigType) as IMixedRealityCameraRig;
-                Debug.Assert(CameraRig != null);
-            }
-
-#if XRTK_USE_LEGACYVR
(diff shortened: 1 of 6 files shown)


XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -156,55 +156,19 @@ public override void Initialize()
...
-                Debug.Assert(CameraRig != null);
-            }
-
-#if XRTK_USE_LEGACYVR
-            ApplySettingsForDefaultHeadHeight();
-#else
-            // We attempt to intialize the camera tracking origin, which might
-            // fail at this point if the subystems are not ready, in which case,
-            // we set a flag to keep trying.
-            trackingOriginInitialized = SetupTrackingOrigin();
```
FejZa, line 177: This was in Initialize because it only needs to happen once. Now that you moved it to OnEnable it will run everytime the service is enabled potentially. It needs to be updated, refactored so that it doesn't run tracking origin initialization if it is already initiailzed.

SimonDarksideJ:
Making it work altered the tests. Looking for the best way to raise the Profile not set exception

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -325,6 +334,7 @@ private bool SetupTrackingOrigin()
             // we can still keep going and assume everything is ready.
             var trackingOriginModeSet = true;
 
+#if !UNITY_EDITOR
```
FejZa, line 337: Why add this unnecessary preprocessor? The previous implemetnation was working perfectly fine without it?

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -345,6 +355,10 @@ private bool SetupTrackingOrigin()
                 // connected, position the camera at the configured default offset.
                 UpdateCameraHeightOffset(defaultHeadHeight);
             }
+#else
```
FejZa, line 358: See comment above. This does nothing except adding useless lines of code.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -435,9 +449,9 @@ protected virtual void ApplySettingsForDefaultHeadHeight()
 #endif
 
 #if !XRTK_USE_LEGACYVR
-        /// <summary>
-        /// Updates the camera height offset to the specified value.
-        /// </summary>
+            /// <summary>
```
FejZa, line 452: Formatting was broken here

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -402,7 +416,7 @@ private void UpdateTrackingOrigin(TrackingOriginModeFlags trackingOriginModeFlag
             SyncRigTransforms();
         }
 
-        #endregion Tracking Origin Setup
```
FejZa, line 405: Formatting was broken here

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
FejZa: This will also break things when working with Meta Link or the mock device plugin.

FejZa (changes requested):

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -311,7 +275,52 @@ public override void Destroy()
             cameraSystem.UnRegisterCameraDataProvider(this);
         }
 
-        #endregion IMixedRealitySerivce Implementation
+#endregion IMixedRealitySerivce Implementation
```
StephenHodgson, line 278: formatting is wrong here

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/LocomotionSystem/LocomotionSystem.cs
```diff
@@ -91,7 +91,10 @@ public override void Disable()
         /// <inheritdoc />
         public override void Destroy()
         {
-            CameraCache.Main.gameObject.EnsureComponentDestroyed<LocomotionProviderEventDriver>();
+            if (Camera.main != null)
+            {
+                CameraCache.Main.gameObject.EnsureComponentDestroyed<LocomotionProviderEventDriver>();
```
StephenHodgson, line 96: not really sure why we're checking for null here, We should likely roll this into the ceched check.

SimonDarksideJ: It can't
Because at this point, if the Camera is null then there is no need to destroy the component.
And if you use the CameraCache, it will create a camera and the reference will still not exist.

StephenHodgson:
Overall I am happy with lifting the camera setup out into their own method calls.

Still needs to pass validation.

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/LocomotionSystem/LocomotionSystem.cs
StephenHodgson: Pretty sure it was checking the same value, as before but maybe lets make a dedicated `CameraCache.IsNull()` or something, and an optional flag to refresh the reference.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: I'll re-evaluate, however, this was raising issues in the editor when you hit play and spamming the console because the tracking type was "unknown"

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: Shouldn't as it only affects the editor itself, but I think I see what you are getting at

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: Fixed

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/LocomotionSystem/LocomotionSystem.cs
SimonDarksideJ: Done, added IsNull and IsNotNull properties to CameraCache

SimonDarksideJ:
**DO NOT MERGE**
Still need to resolve an issue with the MockHMD XR provider not being recognised properly

SimonDarksideJ:
New issue identified, that the toolkit is initialising when the toolkit is not in the scene.

StephenHodgson:
> New issue identified, that the toolkit is initialising when the toolkit is not in the scene

I know for a fact this isn't true, bc I have at least one client who only uses the toolkit for the build pipeline, and it was initializing they would have said something. Unless this PR breaks that behavior and introduces a regression.

SimonDarksideJ:
> > New issue identified, that the toolkit is initialising when the toolkit is not in the scene
> 
> I know for a fact this isn't true, bc I have at least one client who only uses the toolkit for the build pipeline, and it was initializing they would have said something. Unless this PR breaks that behavior and introduces a regression.
As I reported it in this PR, yes it seems to be related.

Or as I've had in the past, Unity is confused :D Hence why I added this as another blocking issue for this PR

SimonDarksideJ:
> New issue identified, possibly due to these updates. The toolkit is initialising when the toolkit is not in the scene. Investigating

Ok, panic over it was a non-issue and just Unity being Unity.   After 3 Scene creations and deleting the XRTK start scene, it stopped being silly.

On a side note, it occurred after a short incident of the XRTK not allowing ply mode because the Profile was not set.  However, it WAS set in the inspector.  Something to keep an eye on.

StephenHodgson:
Are you sure it's a new issue, or are you just breaking things that used to work?

XRTK-Core/Packages/com.xrtk.core/Runtime/Utilities/CameraCache.cs
```diff
@@ -52,5 +52,21 @@ public static Camera Refresh(Camera newMain)
 
             return cachedCamera = newMain;
         }
+
+        /// <summary>
+        /// Validates if the CameraCache reference is Null to avoid creating a new camera if it does not exist.
+        /// </summary>
+        public static bool IsNull
```
StephenHodgson, line 59: Love this change

SimonDarksideJ: Here's hoping that is truthful and not mocking :D

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -156,18 +156,15 @@ public override void Initialize()
...
             base.Initialize();
 
-            if (CameraRig == null)
-            {
-                if (CameraCache.Main.transform.parent.IsNull())
-                {
-                    var rigTransform = new GameObject().transform;
-                    CameraCache.Main.transform.SetParent(rigTransform);
-                }
+            cameraSystem.RegisterCameraDataProvider(this);
```
StephenHodgson, line 159: Certain this is the line that is initializing

StephenHodgson: Maybe moving this down below that is playing check may help

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
```diff
@@ -316,10 +316,11 @@ void OnPlayModeStateChanged(UnityEditor.PlayModeStateChange playModeState)
                 }
 #endif // UNITY_EDITOR
 
-                EnsureMixedRealityRequirements();
-
+                // if the Toolit has a profile, validate toolkit requirements and initialise services
                 if (HasActiveProfile)
                 {
+                    EnsureMixedRealityRequirements();
```
StephenHodgson, line 322: There's no point in putting this within the check.
If you've already gotten here then we've initialized the service locator and this needs to run

SimonDarksideJ:
> Are you sure it's a new issue, or are you just breaking things that used to work?

Non-issue, it was just Unity playing up.  As no code has changed and now it's not happening anymore.

XRTK-Core/Packages/com.xrtk.core/Runtime/Services/MixedRealityToolkit.cs
SimonDarksideJ: Except if there is no profile, then is should not run.  Because it then causes other issues because it is running with no profile

StephenHodgson: See https://github.com/XRTK/com.xrtk.core/pull/919#discussion_r800188587

StephenHodgson:
> Non-issue, it was just Unity playing up. As no code has changed and now it's not happening anymore.

https://github.com/XRTK/com.xrtk.core/pull/919#discussion_r800188587. It's definitely because of some cone changes you made cuz it's not the behavior that was happening before this pull request

SimonDarksideJ:
> > Non-issue, it was just Unity playing up. As no code has changed and now it's not happening anymore.
> 
> [#919 (comment)](https://github.com/XRTK/com.xrtk.core/pull/919#discussion_r800188587). It's definitely because of some code changes you made cuz it's not the behavior that was happening before this pull request

It has also STOPPED happening now, after reloading and restarting Unity.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: The behaviour for this line has not changed.  It was called in initialise before and is still being called in initialise now.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: I've reverted the change and implemented a check for the MockHMD which reports as a system but does not support tracking.  Works as expected now.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: Removed

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: Removed

SimonDarksideJ:
Ok, changes tested and working as expected now with no adverse effects.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -313,6 +310,21 @@ public override void Destroy()
 
         #endregion IMixedRealitySerivce Implementation
 
+        private void EnsureCameraRigSetup()
```
StephenHodgson, line 313: Should we add a menu item to setup rig manually like you are requesting in #911

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -402,7 +420,7 @@ private void UpdateTrackingOrigin(TrackingOriginModeFlags trackingOriginModeFlag
             SyncRigTransforms();
         }
 
-        #endregion Tracking Origin Setup
```
StephenHodgson, line 405: formatting is still broken here.

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -313,6 +310,21 @@ public override void Destroy()
 
         #endregion IMixedRealitySerivce Implementation
 
+        private void EnsureCameraRigSetup()
+        {
+            if (CameraRig == null)
+            {
+                if (CameraCache.Main.transform.parent.IsNull())
+                {
+                    var rigTransform = new GameObject().transform;
```
StephenHodgson, line 319: Probably a good idea to give this object a name

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: We "could" but that would be counter productive at this time, the automatic validation has been retained.

And the MixedRealityToolkit -> Configure option in the menu performs the manual setup already.

StephenHodgson (changes requested):
See outstanding comments

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: I hadn't changed that code, but sure, we can give it the same default as in other places.

StephenHodgson:
also resolving comments without an actual resolution doesn't make them "resolved"

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: Fixed

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
SimonDarksideJ: Added toolkit default for the default camera rig name and applied it to all places referring to or creating the default camera rig

SimonDarksideJ:
Comments addressed

StephenHodgson (approved):

SimonDarksideJ:
I thought the author of the PR was the one supposed to merge once it had been accepted??
To ensure they hadn't had any further thoughts since posting it for review?

XRTK-Core/Packages/com.xrtk.core/Runtime/Providers/CameraSystem/BaseCameraDataProvider.cs
```diff
@@ -245,9 +245,9 @@ public override void Update()
             }
 
 #if !XRTK_USE_LEGACYVR
-            // We keep trying to intiailze the tracking origin,
+            // We keep trying to initialize the tracking origin,
             // until it worked, because at application launch the
-            // subystems might not be ready yet.
+            // subsytems might not be ready yet.
```
SimonDarksideJ, line 250: This is still spelled incorrectly :D
Shouldn't it be subsystems?

Granted I also missed it

XRTK-Core/Packages/com.xrtk.core/Runtime/Utilities/CameraCache.cs
```diff
@@ -54,19 +54,13 @@ public static Camera Refresh(Camera newMain)
...
         /// </summary>
-        public static bool IsNull
-        {
-            get { return Camera.main == null || cachedCamera == null; }
-        }
+        public static bool IsNull => Camera.main == null || cachedCamera == null;
 
         /// <summary>
-        /// Validates if the CameraCache reference is Not Null to avoid creating a new camera if it does not exist.
+        /// Validates if both the CameraCache reference as well as <see cref="Camera.main"/> is NotNull to avoid creating a new camera if it does not exist.
```
SimonDarksideJ, line 62: Nice adding of See refs, missed that.

PR #920, opened 2022-02-09
Reset XR loader configuration as it was producing unintended results

SimonDarksideJ:

## Overview

Disabled and updated Unity XR loader configuration back to original XRTK settings

## Changes

Turns off all XR loaders by default and updates settings based on current layout.  Developers should only enable loaders in development for those they are testing with.


SimonDarksideJ:
@StephenHodgson can you check the error here?
It's complaining about a submodule for WebXR but WebXR is not part of this PR or branch.   Is there a caching issue from the WebXR PR runs?

PR #921, opened 2022-03-18
added the ability to split android apks into expansion files

StephenHodgson:

added the ability to split android apks into expansion files


SimonDarksideJ:
So it is an extension to pass a command-line build argument to set a Unity build setting.  Curious how Unity themseleves do not support this.  Nice work.

